PR #1, opened 2019-05-30
use jackson type resolution

CarstenWickner:
Instead of own type resolution – benefiting from the established (assumed to be more stable) code from `jackson-databind`.

Unfortunately, one limitation is that `Enum` and `Comparable` are "optimised" not to include their respective type parameter. For `Enum` that is assumed to be a self-reference of the enum implementation and can be (somehow) worked around, but for `Comparable` there is no telling what it may be compared with (but is simply assumed to be a self-reference as well).

As of `jackson-databind` version 2.9.9, it is not possible to turn off this premature optimisation. This has been raised here: https://github.com/FasterXML/jackson-databind/issues/2337

For now, the question is:
1. Rather stick with the current type resolution present here (with all the potential issues that might have) or
2. Accept this one known limitation of the type resolution from `jackson-databind` and profit from its proven stability in other respects.


CarstenWickner:
Closing this in favour of #2 due to the aforementioned limitations of the JSON specific type resolution in `jackson-databind`.

PR #4, opened 2019-08-17
Allow passing type parameters on generateSchema

CarstenWickner:
Underlying type resolution allows for a parameterized type to passed-in together with its type parameters.

This is simply being exposed here on the main `SchemaGenerator.generateSchema()` method, while still being backward-compatible, i.e. if no type parameters are provided for a given parameterized type, only the lower bounds of the respective type parameters will be considered (if there are any).


CarstenWickner:
This has been released under version [3.2.0](https://github.com/victools/jsonschema-generator/releases/tag/v3.2.0).

Issue #9, opened 2019-08-31
Javax validation: add support for @Pattern


hleb-albau:
`@Pattern` should be attached to schema as pattern.
See https://json-schema.org/understanding-json-schema/reference/string.html#regular-expressions

CarstenWickner:
HI @hleb-albau,

same as per your other request #8: 
supporting the "pattern" property out-of-the-box could be added as a separate feature in the `jsonschema-generator`.
Adding support for any particular `@Pattern` annotation could then be added via configuration (either directly or by providing your own module implementation).

For now, you can achieve this by providing an `InstanceAttributeOverride` via:
- `SchemaGeneratorConfigBuilder.forFields().withInstanceAttributeOverride(override)` or
- `SchemaGeneratorConfigBuilder.forMethods().withInstanceAttributeOverride(override)`

Since it can already be achieved, I won't consider this a high priority but will look into it.

CarstenWickner:
HI @hleb-albau,

I'm happy to announce that the general support for the `"pattern"` property has been implemented and released under version [v3.2.0](https://github.com/victools/jsonschema-generator/releases/tag/v3.2.0) of the core `victools/jsonschema-generator` library.
The specific support of a regular expression being looked-up from `@Pattern` or optionally from `@Email` was implemented and released under version [v3.2.0](https://github.com/victools/jsonschema-module-javax-validation/releases/tag/v3.2.0) of the `victools/jsonschema-module-javax-validation`.

Please be aware, that the current implementation does **NOT** take any of the additional `flags` into account that may be provided on `@Pattern` or `@Email`. Only the `regexp` is being passed on to the JSON Schema here.
Due to this limitation, I've disabled this feature by default. That makes it necessary to enable the new feature on the `JavaxValidationModule` explicitly by providing `JavaxValidationOption.INCLUDE_PATTERN_EXPRESSIONS` in its constructor.

The usage would look like this then:
```java
ObjectMapper objectMapper = new ObjectMapper();
JavaxValidationModule module = new JavaxValidationModule(JavaxValidationOption.INCLUDE_PATTERN_EXPRESSIONS);
SchemaGeneratorConfigBuilder configBuilder = new SchemaGeneratorConfigBuilder(objectMapper, OptionPreset.PLAIN_JSON)
    .with(module);
SchemaGeneratorConfig config = configBuilder.build();
SchemaGenerator generator = new SchemaGenerator(config);
[...]

PR #5, opened 2019-08-30
Add new Json schema properties

imIfOu:
feat: add "default" value properties
feat: add "required" properties
feat: add resolve metadata
conf: add more rules in gitignore

PR #10, opened 2019-08-31
Additional Property Support: "required", "default", "pattern"

CarstenWickner:
This includes changes proposed by @FlorianDidier in PR #5 (adding support for "required" and "default"). Thanks again!

I've made slight adjustments to not limit this to `String` default values but also support numeric and boolean ones or even any kind of object (the latter is only intended for edge cases and being dependent on a respective configuration of the used `ObjectMapper` to still produce a valid JSON Schema).

Additionally, issue #9 was logged at pretty much the same time with a similar request: supporting the "pattern" property. This has been included here now as well.


CarstenWickner:
This has been released under version [3.2.0](https://github.com/victools/jsonschema-generator/releases/tag/v3.2.0).

PR #17, opened 2020-01-01
Extend custom definitions

CarstenWickner:
Main purpose of these changes:
- Allowing a `CustomDefinitionProviderV2` (replacing `CustomDefinitionProvider`) to hand-off the generation of parts of the custom definition to the normal generator instance

To make this manageable, some logic was moved out of the main `SchemaGenerator` class into the `SchemaGenerationContextImpl` class (renamed from `SchemaGenerationContext` to avoid name conflicts with the newly introduced `SchemaGenerationContext` interface).

The tests were kept on the `SchemaGenerator` level to reduce the amount of work necessary here and to show that the previous behavior remained unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `CustomDefinitionProviderV2` with access to `SchemaGenerationContext`, e.g. to allow continuing normal schema generation for nested properties
+
+### Deprecated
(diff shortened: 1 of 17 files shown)


CarstenWickner:
Added a test to show-case the now possible custom collection definition, e.g.
```java
CustomDefinitionProviderV2 customDefinitionProvider = (javaType, context) -> {
    if (!javaType.isInstanceOf(Collection.class)) {
        return null;
    }
    ResolvedType generic = context.getTypeContext().getContainerItemType(javaType);
    return new CustomDefinition(objectMapper.createObjectNode()
            .put(SchemaConstants.TAG_TYPE, SchemaConstants.TAG_TYPE_OBJECT)
            .set(SchemaConstants.TAG_PROPERTIES, objectMapper.createObjectNode()
                    .set("stream().findFirst().orElse(null)", context.makeNullable(context.createDefinition(generic)))));
};
```

PR #18, opened 2020-01-03
Context independent attributes

CarstenWickner:


CarstenWickner:
Hi @imIfOu,

just fyi, in case it's relevant to you and your [jsonschema-module-addon](https://github.com/imIfOu/jsonschema-module-addon):
I've extended the available configuration options by also allowing schema attributes to be picked-up independent of a particular field or method via this new `SchemaGeneratorConfigBuilder.forTypesInGeneral()` feature.

That way, it is also possible to get a properly populated schema even for items inside collections/arrays (e.g. with "title", "description", etc.) – unfortunately, I had overlooked these before.

Issue #22, opened 2020-02-05
Add "additionalProperties" : false to objects by default


ALEUT:
It would be useful to add `"additionalProperties" : false` for objects (`"type" : "object"`) because by default `additionalProperties` is true and it means, that this object can contain other fields. But if the source of the schema is Java files, it's impossible to have other fields.

Currently I use such a code as workaround:
```java
private static final String TAG_ADDITIONAL_PROPERTIES = "additionalProperties";

.with((jsonSchemaTypeNode, scope, config) -> {
	JsonNode type = jsonSchemaTypeNode.get(TAG_TYPE);
	if (type != null && type.isTextual() && type.textValue().equals(TAG_TYPE_OBJECT)) {
		if (jsonSchemaTypeNode.has(TAG_PROPERTIES)) {
			jsonSchemaTypeNode.put(TAG_ADDITIONAL_PROPERTIES, false);
		}
	}
})
```

Or, as an option, have it configurable using `Option`.

CarstenWickner:
I could also imagine that additional properties are allowed, but will simply be ignored by the Java implementation.

Making this a standard `Option` sounds very reasonable.
You’re welcome to create a PR. Otherwise, I might have a look during the next couple of days myself.

ALEUT:
@CarstenWickner I will be overloaded in nearest week, it would be great it you will look into it.

CarstenWickner:
Hi @ALEUT,

I had a first look into this now and am considering to add this alongside the other fields via its dedicated resolvers in order to allow for things like the Jackson annotation `@JsonIgnoreProperties(ignoreUnknown = true)` to be considered per type instead of just treating this as a single global option. That would still allow you to easily disable it for all types.

However, according to the JSON Schema specification:
> The `additionalProperties` keyword may be either a boolean or an object.
> If `additionalProperties` is an object, that object is a schema that will be used to validate any additional properties not listed in `properties`.

Only catering for the `boolean` side of things seems also rather limited and hard to extend later without a breaking change.
I don't have a nice idea yet how to allow this kind of configuration with type "Boolean | Schema | null"  in a clean but convenient way.

I'm open to suggestions. :)

ALEUT:
Hi @CarstenWickner 
Can it be optional Option then in this case?

CarstenWickner:
Hi @ALEUT,

All `Option`s are optional, but I get what you mean.

However, that wouldn’t solve the issue of having to:
[...]

PR #28, opened 2020-02-21
Support additional properties

CarstenWickner:
This implements feature request #22 – with the extension of not only catering for `false` to be set as `additionalProperties` but also to define a sub-schema for those `additionalProperties`.

Expected usage for the use-case described in #22 could be something like this (depending on what is being treated as `type: "object"` according to other configurations):
```java
SchemaGeneratorConfigBuilder.forTypesInGeneral()
    .withAdditionalPropertiesResolver(scope ->
        scope.isContainerType()
        || scope.getType().isInstanceOf(Number.class)
        || scope.getType().isInstanceOf(String.class)
        || scope.getType().isPrimitive()
            ? null
            : Void.class);
```
- Providing `null` is reserved for the "neutral" behaviour, i.e. not adding an `additionalProperties` attribute and thereby allowing additional properties of any type to be included in a respective JSON instance.
- Providing `Void.class` or `Void.TYPE` results in `"additionalProperties": false` to be included in the generated JSON Schema.
- Providing any other `Type` (can be a `Class`, `GenericType` or `ResolvedType` – e.g. as created via the `scope.getContext().resolve(mainType, typeAttributes...)`) will result in a corresponding sub-schema to be included, e.g. `"additionalProperties": { "type": "string" }`

@ALEUT what do you reckon? Does that fulfill your requirements?
The part about having to implicitly filter-out non-"object" schemas (where `additionalProperties` wouldn't make any sense) is not too nice, but I'm not sure how properly handle this in a more generic way.


ALEUT:
Hi @CarstenWickner 
Looks slightly tricky to use (looks it's enough difficult to apply the resolver without an example). What will be for Maps, for which we just have `type: object` and nothing else. It will be object, but shouldn't have "additionalProperties": false.

CarstenWickner:
Hi @ALEUT,
to explicitly allow `additionalProperties`, the resolver would simply have to return `Object.class`.

CarstenWickner:
Merging this as-is, since I believe that already to be a meaningful extension of the library.

Not closing #22 though, since I still got some ideas to make the usage more convenient/intuitive.

ALEUT:
@CarstenWickner could you please show what will be generated for any map? Just `type: object` without `additionalProperties`?

CarstenWickner:
Hi @ALEUT,

You have possibilities:
- simply omitting the `additionalProperties`, or
- specifying the `Map` defined value type as sub-schema.

Please refer to the newly added example in the README for that particular use-case and let me know if that’s not clear enough.

PR #32, opened 2020-02-24
feat: default for additional/patternProperties

CarstenWickner:
Changes made in the interest of making the usage of `additionalProperties`/`patternProperties` more convenient (as per #22):
- Specifically omit both newly supported keywords in case of the custom definitions for simple types (i.e. the non-`object` ones as well as the base `Object.class`).
- Introduce new `Option.FORBIDDEN_ADDITIONAL_PROPERTIES_BY_DEFAULT` that applies to all definitions except for the specifically ones mentioned above and anything with `TypeScope.isContainerType()` being `true`.

### DIFF
--- a/src/test/java/com/github/victools/jsonschema/generator/impl/SchemaGeneratorConfigImplTest.java
+++ b/src/test/java/com/github/victools/jsonschema/generator/impl/SchemaGeneratorConfigImplTest.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package com.github.victools.jsonschema.generator.impl;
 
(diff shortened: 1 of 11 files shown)


src/test/java/com/github/victools/jsonschema/generator/impl/SchemaGeneratorConfigImplTest.java
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.github.victools.jsonschema.generator.impl;
```
CarstenWickner: No reason to remove this empty line.
```suggestion

package com.github.victools.jsonschema.generator.impl;
```

PR #58, opened 2020-03-27
Jackson subtype resolution

CarstenWickner:
Introducing subtype lookup in Jackson module:
- [x] Supporting `@JsonTypeInfo` and `@JsonSubTypes` on supertype
    - [x] Considering `@JsonTypeInfo.include` with values `As.PROPERTY`, `As.EXISTING_PROPERTY`, `As.WRAPPER_ARRAY`, `As.WRAPPER_OBJECT`
    - [x] Considering `@JsonTypeInfo.use` with values `Id.CLASS`, `Id.NAME` (expecting `@JsonTypeName` on subtype)
- [x] Supporting `@JsonTypeInfo` and `@JsonSubTypes` on properties as override of the per-type behaviour
    - [x] Considering `@JsonTypeInfo.include` with values `As.PROPERTY`, `As.EXISTING_PROPERTY`, `As.WRAPPER_ARRAY`, `As.WRAPPER_OBJECT`
    - [x] Considering `@JsonTypeInfo.use` with values `Id.CLASS`, `Id.NAME` (expecting `@JsonTypeName` on subtype)
    - [x] Applying per-property override even if field/method is a container type (i.e. array or collection)
- [x] Include unit tests for various combinations
- [x] extend module integration tests to also cover container items

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Fixed
 - Should consider per-type attributes even on inline custom definitions
-- Look-up per-type subtypes for properties if there is no per-property target type override
 - Use less strict `anyOf` instead of `oneOf` when indicating that a sub-schema may be of `"type": "null"`
 
 #### Dependency Update
 - `com.fasterxml.jackson.core`:`jackson-core`/`jackson-databind` from `2.10.2` to `2.10.3`
-- removed dependencies to `log4j` implementation (only `slf4j-api` remains)
+- Remove dependencies to `log4j` implementation (only `slf4j-api` remains)
 
 ### `jsonschema-module-jackson`
 #### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for custom definitions in the scope of a particular field/method via `SchemaGeneratorConfigPart.withCustomDefinitionProvider()`
 - Ability to opt-out of normal "attribute collection" for custom definitions through new `CustomDefinition` constructor parameters
 
+#### Changed
+- If a field/method is of a container type: apply the per-property configurations on its items; with `MemberScope.isFakeContainerItemScope()` flag
+
 #### Fixed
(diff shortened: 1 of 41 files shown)


CarstenWickner:
Hi @JanLabrie,

You've mentioned before that you already handled the subtype resolution in your fork of the `jsonschema-module-jackson`.
If you have time, I'd appreciate if you could have a look at my approach here and let me know whether that matches your expectations in regards to the `WRAPPER_OBJECT` scenario at least.

There are two open points here for which I need to make a decision how to go about them:
1. When there is a `@JsonTypeInfo` annotation on an array/collection property that overrides the corresponding annotation on the collection item's supertype. Since that might get complicated, I will probably leave this out now until someone actually uses this and has an issue with it not being supported here.
2. When a subtype is directly declared on a property, would the same `WRAPPER_OBJECT` handling still apply, even though there is technically no need for identifying which subtype it is?
Right now, I assume that this serialization strategy is always applied, but I've never used this before and didn't yet test how jackson behaves.
I guess I should do that. But your input is appreciated anyway. 😉

And as per your comment on the other issue: after this PR is merged, I'm planning to release version 4.8.0 of the main generator and all (standard) modules and then archive their "standalone" repositories in favor of the maven multi-module project here.

CarstenWickner:
Just for completeness' sake, to document my decisions for the aforementioned two points:
1. When there is a `@JsonTypeInfo` annotation on an array/collection property that overrides the corresponding annotation on the collection item's supertype.
I ended up including this through a new `MemberScope.isFakeContainerItemScope()` flag in the main generator since that seemed like a common use-case: to apply some field/method annotations not to the declared `List<X>` but rather to the single `X` container item.
2. When a subtype is directly declared on a property, the same `WRAPPER_OBJECT` handling is being applied by Jackson even though there is technically no need for identifying which subtype it is.
I confirmed this assumption by actually performing a JSON Schema validation based on my generated schema (in an integration test) against an example JSON instance (produced with Jackson).

The changes are quite extensive now (appearing even more so due to the added tests) and may lead to different schema output depending on the extent of user configurations. Since the (standard) modules are now aligned in their release cycle, at least those are already adjusted to consider the `MemberScope.isFakeContainerItemScope()` flag accordingly, e.g. for a `@Size(min = 2) List<String>` the annotation should still only apply to the `List` and not the `String`.

PR #78, opened 2020-04-21
First version maven plugin

JanLabrie:
Hi Carsten,

This is a branch that contains the first version of the Maven plugin I wrote for the JSON generator.

Please have a look at it and merge when it is satisfactory.

Thank you,

Jan


JanLabrie:
Premature PR. Fixing the unit test now.

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+<configuration>
+    <className>com.myOrg.myApp.myclass</className>
+    <schemaFileName>mySchema.json</schemaFileName>
+    <schemaFilePath>src/main/resources/schemas</schemaFilePath>
+</configuration>
+```
+- The default name of the file is the name of the class extended with ".schema.json"
+- The default path is "src/main/resources"
+### Configuring schema version
+The version of JSON Schema that is to be used can be configed with the schemaVersion element.
```
CarstenWickner, line 45: typo 😉
```suggestion
The version of JSON Schema that is to be used can be configured with the `<schemaVersion>` element.
```

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+```xml
+<configuration>
+    <className>com.github.victools.jsonschema.plugin.maven.TestClass</className>
```
CarstenWickner, line 48: Maybe stick to the more generic example class, to avoid any ambiguity:
```suggestion
    <className>com.myOrg.myApp.MyClass</className>
```

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+    </configuration>
+</plugin>
+```
+This will use the default configuration of the generator. This will enable all of the modules that are part of the generator. 
+
+
+### Configuring file locations
+```xml
+<configuration>
+    <className>com.myOrg.myApp.myclass</className>
```
CarstenWickner, line 37: Let's follow standard naming conventions also in those examples (same applies for the others below). 😄
```suggestion
    <className>com.myOrg.myApp.MyClass</className>
```

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+    <schemaVersion>DRAFT_2019_09</schemaVersion>
+</configuration>
+```
+Allowed values are:
+- DRAFT_2019_09
+- DRAFT_7
+
+The last one is the default
+### Configuring modules
+When you want to have tighter control over the modules that are to be used during generation, the <module> element can be used to define them
```
CarstenWickner, line 58:
```suggestion
When you want to have more control over the modules that are to be used during generation, the `<module>` element can be used to define them.
```

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+                <goal>schema-generator</goal>
+            </goals>
+        </execution>
+    </executions>
+    <configuration>
+        <className>com.myOrg.myApp.myclass</className>
+    </configuration>
+</plugin>
+```
+This will use the default configuration of the generator. This will enable all of the modules that are part of the generator.
```
CarstenWickner, line 31: The issue with this default behavior is, that the modules expect certain `<scope>provided</scope>` dependencies. Isn't that failing at runtime then, if those are not present?

A more conservative default behavior could be to not use any modules and rather expect them to be configured explicitly – depending on what is actually present.

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+```xml
+<configuration>
+    <className>com.myOrg.myApp.myclass</className>
+    <schemaFileName>mySchema.json</schemaFileName>
+    <schemaFilePath>src/main/resources/schemas</schemaFilePath>
+</configuration>
+```
+- The default name of the file is the name of the class extended with ".schema.json"
+- The default path is "src/main/resources"
+### Configuring schema version
```
CarstenWickner, line 44: Maybe add an empty line above each subsection header (also the following ones):
```suggestion

### Configuring schema version
```

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+<configuration>
+    <className>com.github.victools.jsonschema.plugin.maven.TestClass</className>
+    <schemaVersion>DRAFT_2019_09</schemaVersion>
+</configuration>
+```
+Allowed values are:
+- DRAFT_2019_09
+- DRAFT_7
+
+The last one is the default
```
CarstenWickner, line 56: I recently added `DRAFT_6` as well, since there was hardly any relevant difference to `DRAFT_7`.
```suggestion
Allowed values are:
- `DRAFT_2019_09`
- `DRAFT_7`
- `DRAFT_6`

`DRAFT_7` is the default, if not specified.

```

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+```
+in this way the options as supported by the module can be specified.
+### Defining the use of custom modules
+To enable a custom module in the generation the following construct can be used:
+```xml
+<configuration>
+    <className>com.myOrg.myApp.myclass</className>
+    <modules>
+        <module>
+            <name>MySchemaModule</name>
```
CarstenWickner, line 92: Why is the `<name>` required in addition to the `<className>`?
Also slightly confusing that the name does not match the class name in this example.

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+            <className>com.myOrg.myApp.MyAddonModule</className>
+        </module>
+    </modules>
+</configuration>  
+```
+Make sure your custom module is on the classpath.
+Setting options on custom modules is not yet supported.
```
CarstenWickner, line 99: I assume the expectation is also that the identified module class has a no-args constructor?
I'd be hesitant to include "not yet" here, since it may never be added and we wouldn't want to create unrealistic expectations. 😉

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+            <artifactId>jsonschema-maven-plugin</artifactId>
+            <version>4.8.0</version>
+            <configuration>
+                <className>com.github.victools.jsonschema.plugin.maven.TestClass</className>
+                <schemaVersion>DRAFT_2019_09</schemaVersion>
+                <schemaFileName>testclass.json</schemaFileName>
+                <schemaFilePath>result/schemas</schemaFilePath>
+                <options>
+                    <option>DEFINITIONS_FOR_ALL_OBJECTS</option>
+                    <option>OTHER_OPTION</option>
```
CarstenWickner, line 115: Let's pick an existing second option for this example.
And I assume the `OptionPreset.PLAIN_JSON` is being used by default? That should be the most common I guess, or should it be possible to configure that preset as well (to avoid a long list of individual options)?

Since the `OptionPreset` is currently fixed, how would one go about removing an `Option`? E.g. if you don't want the `Option.SCHEMA_VERSION_INDICATOR` to be included?

JanLabrie:
Unit test fixed.

jsonschema-maven-plugin/pom.xml
```diff
@@ -0,0 +1,271 @@
...
+        <dependency>
+            <groupId>org.apache.maven.plugin-testing</groupId>
+            <artifactId>maven-plugin-testing-harness</artifactId>
+            <version>3.3.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.10.3</version>
```
CarstenWickner, line 67: This `pom.xml` can be aligned a bit more with the others (i.e. moving some things into the parent `pom` and reducing the duplication here).

But I can also do that after this has been merged.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -0,0 +1,415 @@
...
+public class SchemaGeneratorMojo extends AbstractMojo {
+
+    @Parameter(defaultValue = "${project}", required = true, readonly = true)
+    MavenProject project;
+
+    /**
+     * Full name of the class for which the JSON schema will be generated.
+     */
+    @Parameter(property = "className", required = true)
+    private String className;
```
CarstenWickner, line 72: What changes would be necessary to allow for multiple schemas to be generated?

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -0,0 +1,415 @@
...
+        // Add options when required
+        if (options.length == 0) {
+            setDefaultOptions(configBuilder);
+        } else {
+            setOptionsFromConfig(configBuilder, options);
+        }
+
+        // Register the modules
+        if (modules.length == 0) {
+            configureDefaultModules(configBuilder);
```
CarstenWickner, line 218: As per other comment, I'd suggest to rather not include any modules by default.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -0,0 +1,415 @@
...
+                    getLog().info("- Adding Javax Validation Module");
+                    addJavaxValidationModule(configBuilder, module);
+                    break;
+                case "Swagger":
+                    getLog().info("- Adding SwaggerModule");
+                    addSwaggerModule(configBuilder, module);
+                    break;
+                default:
+                    throw new MojoExecutionException("Error: Module does not have a name in "
+                            + "['Jackson', 'Validation', 'Swagger'] or does not have a custom classname.");
```
CarstenWickner, line 306: `'JavaxValidation'` != `'Validation'` 😉

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -0,0 +1,415 @@
...
+                switch (module.name) {
+                case "Jackson":
+                    getLog().info("- Adding JacksonModule");
+                    addJacksonModule(configBuilder, module);
+                    break;
+                case "JavaxValidation":
+                    getLog().info("- Adding Javax Validation Module");
+                    addJavaxValidationModule(configBuilder, module);
+                    break;
+                case "Swagger":
```
CarstenWickner, line 300: There is a big difference between Swagger 1.5.x (for which there is a module already) and Swagger 2.x (now the OpenAPI standard, for which there is no module yet but will most likely come in the future).

Please include the 1.5 here as well, to avoid breaking or confusing changes when the other Swagger module is being added.

jsonschema-maven-plugin/src/test/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojoTest.java
```diff
@@ -0,0 +1,94 @@
...
+        Xpp3Dom pomDom = Xpp3DomBuilder.build(new StringReader(pom));
+        pluginConfiguration = extractPluginConfiguration("jsonschema-maven-plugin", pomDom);
+    }
+
+    /**
+     * Unit test to test the basic working with a complete configuration
+     *
+     * @throws Exception In case something goes wrong
+     */
+    public void testBasicWorking() throws Exception {
```
CarstenWickner, line 84: Shouldn't there be a `@Test` annotation here?

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
...
+### Plugin definition
+```xml
+<plugin>
+    <groupId>com.github.victools</groupId>
+    <artifactId>jsonschema-maven-plugin</artifactId>
+    <version>4.8.0</version>
+    <executions>
+        <execution>
+            <goals>
+                <goal>schema-generator</goal>
```
CarstenWickner, line 22: Maybe `generate` or `generate-schema` are nicer names for a `goal`?

How does this work exactly? When is the schema being generated? During each build (since some contained class might have been changed)? Or during what phase is this triggered by default?
I assume the phase can be configured, but can this task be skipped explicitly via some argument, e.g. `schema.skip` or something like that?

CarstenWickner:
Hi Jan,

Thanks a lot for that. Great that you got the permission to contribute this.

Overall, I'd say it looks good. I'd suggest a couple of small changes though.

----

How do you wanna play this?
- Shall I review/comment and you do the changes yourself?
- Otherwise, I'd merge this as-is on a feature branch, make the changes myself and then merge it to the `master` branch to be included in the next release (4.11.0) then.

JanLabrie:
Hi Carsten,

I propose that I do the changes on the branch and we take it to a level
where it is good enough to be merged.

Jan

Op di 21 apr. 2020 om 14:53 schreef Carsten Wickner <
notifications@github.com>:

jsonschema-maven-plugin/README.md
```diff
@@ -0,0 +1,133 @@
+# Java JSON Schema Generator – Maven Plugin
+[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.github.victools/jsonschema-module-javax-validation/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.github.victools/jsonschema-maven-plugin)
```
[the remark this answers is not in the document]
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: True. Is better.

jsonschema-maven-plugin/README.md
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: Actually <name> is not used when <className> is set. So I will remove it.

jsonschema-maven-plugin/README.md
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: - Changed to NULLABLE_FIELDS_BY_DEFAULT. Is that good one?
- OptionPreset.PLAIN_JSON is indeed used by default. This is currently not configurable.
- The current approach is that the default is OptionPreset.PLAIN_JSON, and all the options from the <options> are added the configBuilder using the with() method. without() is currently not supported.
Let's have a discussion how we can add this feature.

jsonschema-maven-plugin/pom.xml
JanLabrie: 'But I can also do that after this has been merged.'....Yes please

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
JanLabrie: - We could make the className a collection type, so you can specify multiple classes.
In that case we need to think about a naming convention for the generated schema files when the default file name is not good enough.
What is actually a good naming convention for JSON schema files?
- There was also a request to use all classes in a package/folder. We would need to investigate how we can find all classes in a package/folder.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
JanLabrie: ok

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
JanLabrie: ok

jsonschema-maven-plugin/README.md
JanLabrie: - True. Changed goal to `generate-schema`
- Currently the generation is done in the `PROCESS_TEST_SOURCES` phase. I am not very sure about that. I selected it because is the first phase after the `COMPILE` phase in the phase list for a `jar`.
In case you want to skip the generation of the schema I propose to use the Maven profiling mechanism. In this setup you put the plugin definition is a profiled section.
This is standard Maven technology. I could spend some words on it in the Readme. Let's think about this.

jsonschema-maven-plugin/src/test/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojoTest.java
JanLabrie: A `@Test` annotation would be more modern. I will work on the unit test a little more. On my list are:
- JUnit 4 annotations
- More unit tests with better coverage of the functionality.

For now I propose to merge this branch.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
JanLabrie: The main question is whether we can identify a default setup that would be sufficient for a majority of the users. My take is take configuring all modules does not hurt when you do not use the association annotations in your code. But you are right you need to have all the modules in your dependencies.

In case there is no good default behavior...let's kick it out.

I have not done that yet.

jsonschema-maven-plugin/README.md
JanLabrie: See comment on the later issue on the same topic.

jsonschema-maven-plugin/README.md
JanLabrie: Maybe merge what we have now and create an issue to resolve this.

JanLabrie:
I have commented and resolved all review feedback now. Please re-review.

jsonschema-maven-plugin/README.md
CarstenWickner: - The `Option.NULLABLE_FIELDS_BY_DEFAULT` sounds alright for this example, yes.
- Fair enough to tackle the preset topic separately. Maybe we could also keep it simple here and only have a `boolean` configuration flag for disabling any preset to be applyied (i.e. an empty `new OptionPreset()` instead). That way, full control is possible (although with a long list of individual options) and the default behavior can still be `OptionPreset.PLAIN_JSON`.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
CarstenWickner: - Having both options side-by-side (a single `className` and a `classNames` collection) sounds good to me.
- I've no idea what a common naming convention for schema files is. Your proposed `*.schema.json` sounds good enough to me (except for maybe an underscore before the `schema` part? `*_schema.json`).
- As to the file name configuration, maybe we could have a nested structure then:
```xml
<configuration>
    <targetTypes>
        <targetType>
            <className>a.b.c.YourClass</className>
        </targetType>
        <targetType>
            <className>a.b.c.MyClass</className>
            <schemaFileName>MyClass_schema.json</schemaFileName>
        </targetType>
    </targetTypes>
    <targetPackages>
        <targetPackage>x.y.z</targetPackage>
    </targetPackages>
    <schemaFilePath>src/main/resources/schema</schemaFilePath>
</configuration>
```
- I've been using `org.reflections:reflections` in the past for things like looking-up all classes from a certain package. It's easy enough to use and introducing an extra dependency shouldn't be an issue for this maven plugin. I'd also find it acceptable if the file names cannot be configured for this kind of per-package schema generation.

jsonschema-maven-plugin/README.md
JanLabrie: I have moved it to the Compile phase now. Let's assume that the java compilation will always be the first step in that phase.

JanLabrie:
Ok. I will create a new pull request to merge the current status into your new `maven-plugin` branch.
Work items we have left are:
- Options specification (base, enable, disable)
- Schemas for multiple classes
- Unit test improvements
- Discussion on the default configuration
I'll create issues for these items (when not already there)

PR #88, opened 2020-04-26
Maven Plugin cleanups

CarstenWickner:
@JanLabrie I've made a couple small changes to the pending Maven Plugin.

Since it is your creation, I'd appreciate if you could have a look – I don't want to "just merge" it without asking.

 1. renamed goal to `generate` to make the command-line call nicer. that is now `jsonschema:generate` (unless you think `jsonschema:generate-schema` is better?)
 2. update pom to avoid unnecessary plugin management (that is being taken care of by the parent pom already)
 3. some small fixes in README
 4. align maven plugin code with existing codebase (import order + using `this.` prefix)
 5. getting rid of the explicit `ClassLoader` code by adding dependency scope on `@Mojo` annotation
 6. stricter configuration of `SchemaVersion` and `OptionPreset` through enums (therefore no need to specifically map values anymore)
 7. better error messages for invalid module options (it was returning `null` instead of the configured name)
 8. create schema parent directories if necessary (mostly relevant for first-time executions, but I liked the convenience)
 9. close `FileOutputStream` after writing schema to file (using `try-with-resources` which calls `close()` on IO resources automatically)
 10. Changing the log messages regarding the configuration parameters to `debug` level.

----

The schema generation for multiple types is still outstanding though. I didn't touch that area since you were planning to tackle that yourself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ This project consists of:
     - [victools/jsonschema-module-jackson](jsonschema-module-jackson) – deriving JSON Schema attributes from `jackson` annotations (e.g. "description", property name overrides, what properties to ignore) as well as looking up appropriate (annotated) subtypes
     - [victools/jsonschema-module-javax-validation](jsonschema-module-javax-validation) – deriving JSON Schema attributes from `javax.validation` annotations (e.g. which properties are nullable or not, their "minimum"/"maximum", "minItems"/"maxItems", "minLength"/"maxLength")
     - [victools/jsonschema-module-swagger-1.5](jsonschema-module-swagger-1.5) – deriving JSON Schema attributes from `swagger` (1.5.x) annotations (e.g. "description", property name overrides, what properties to ignore, their "minimum"/"maximum", "const"/"enum")
+- the [victools/jsonschema-maven-plugin](jsonschema-maven-plugin) – allowing you to generate JSON Schemas as part of your Maven build
 
 Another example for such a module is:
 - [imIfOu/jsonschema-module-addon](https://github.com/imIfOu/jsonschema-module-addon) – deriving JSON Schema attributes from a custom annotation with various parameters, which is part of the module.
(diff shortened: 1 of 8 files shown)


JanLabrie:
Hi Carsten,

Really good that you made the changes. Please merge.

I am working on the multi-file generation. Finished it, except for the complete package generation.
I made a commit for it and create a PR after yours is merged.

On the full package generation my direction is to use Google Guave as mentioned in the second answer to: [can-you-find-all-classes-in-a-package-using-reflection](https://stackoverflow.com/questions/520328/can-you-find-all-classes-in-a-package-using-reflection)

Is that an OK dependency?

Jan

CarstenWickner:
Hi @JanLabrie,

Merged it now.
Google Guava is fine as a dependency. Maybe a bit big for this use-case and I personally used the Reflections library instead (as per the higher voted answers on that SO question).

But since this is “only” within the maven plugin I’m not too concerned about size.
If you prefer Guava, feel free to use it. I’d just be interested in why you favor that one.

JanLabrie:
Hi Carsten,

I looked at the Reflection library. And what I understood it uses Guava
under the hood. I that case using Guava would be the lightest solution.

 I'll investigate a little more.

Jan

Op ma 27 apr. 2020 09:54 schreef Carsten Wickner <notifications@github.com>:

CarstenWickner:
Hi Jan,

It doesn't seem to be the case anymore that the Reflections library uses Guava internally.
For [version 0.9.11](https://mvnrepository.com/artifact/org.reflections/reflections/0.9.11) it is still being listed, yes.
But not for [version 0.9.12](https://mvnrepository.com/artifact/org.reflections/reflections/0.9.12) anymore (released January 2020).

I'd assume that it is (now) the more lightweight solution. But as said before: this is only the maven plugin, I'm not too concerned about the size of dependencies – more about getting the job done and keeping it maintainable. 😉

Issue #92, opened 2020-04-29
Keep original order of JSON properties


udowal:
Hi,

in my project I need to keep the original order of JSON properties.
In "SchemaGenerationContextImpl.generateObjectDefinition (ResolvedType, ObjectNode)" the 2 lines

```java
final Map<String, JsonNode> targetFields = new TreeMap<>();
final Map<String, JsonNode> targetMethods = new TreeMap<>();
```

could be replaced by something like this:

```java
final Map<String, JsonNode> targetFields;
final Map<String, JsonNode> targetMethods;
if (this.generatorConfig.shouldKeepOriginalPropertiesOrder()) {
    targetFields = new LinkedHashMap<>();
    targetMethods = new LinkedHashMap<>();
} else {
    targetFields = new TreeMap<>();
    targetMethods = new TreeMap<>();
}
```

with the new method `shouldKeepOriginalPropertiesOrder()` and a new appropriate option.

Is it possible?

Best regards
Udo

CarstenWickner:
Hi Udo,

The look-up of fields (and methods) is done via the `com.fasterxml:classmate` library. For your proposal to work we'd have to assume that `classmate` is returning the properties in their declaration order.

If that assumption holds true, then your solution is pretty much all there is to it and it's thereby certainly possible.
Have you tested it already in a fork to confirm that it works as you expect it to?

udowal:
Hi Carsten,

thank you for the fast answer.
I tested it locally without the if-else-statement:

```java
final Map<String, JsonNode> targetFields = new LinkedHashMap<>();
final Map<String, JsonNode> targetMethods = new LinkedHashMap<>();
```

... and it works fine as expected, unit tests are green :).

CarstenWickner:
Hi Udo,

I‘m not concerned about the unit tests, since they don’t check for the order of anything.

I was more interested in whether the properties are actually in the declaration  order then and not some random or still alphabetic order.

udowal:
Hi Carsten,

yes, they are in the declaration order.

CarstenWickner:
Hi Udo,

I tried it now myself but can unfortunately not confirm your observation that the declaration order is being preserved.
Looks to me like `classmate` is not respecting the declaration order – resulting in an unreliable order of the properties being returned – which is the reason why I sorted them alphabetically in the past, to make it at least predictable.

I'm open to alternative suggestions. 😉

udowal:
Hi Carsten,

[...]

PR #93, opened 2020-04-29
Option for properties in declaration order

CarstenWickner:
Implementation of GH-92.

### DIFF
--- a/jsonschema-generator/src/test/java/com/github/victools/jsonschema/generator/SchemaGeneratorOptionsTest.java
+++ b/jsonschema-generator/src/test/java/com/github/victools/jsonschema/generator/SchemaGeneratorOptionsTest.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2020 VicTools.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


jsonschema-generator/src/test/java/com/github/victools/jsonschema/generator/SchemaGeneratorOptionsTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Object parametersForTestGenerateSchema() {
+        return new Object[][]{
+            {"properties in alphabetic order", null, Arrays.asList(
+                "CONSTANT_A", "CONSTANT_B", "fieldInSupertype", "firstField", "lastField", "middleField",
+                "firstMethod()", "lastMethod()", "methodInSupertype()", "middleMethod()"
+                )},
+            {"properties in declaration order", Option.PROPERTIES_IN_DECLARATION_ORDER, Arrays.asList(
+                "fieldInSupertype", "firstField", "middleField", "lastField", "CONSTANT_B", "CONSTANT_A",
+                "middleMethod()", "firstMethod()", "lastMethod()", "methodInSupertype()"
+                )}
```
CarstenWickner, line 46: @udowal this is the order returned by the tests here – unfortunately that is not the declaration order.
Now I remember that I wanted to offer this option as well in the past but didn't manage to convince `classmate` to respect the declaration order.

I'm afraid the change is not as simple as hoped.

PR #106, opened 2020-06-02
- Improved error message in case an empty or incorrect package is given.

JanLabrie:
Hi Carsten,

A small improvement in trying to get a better understandable error message.

Please review, because the code is a little fishy. Maybe we should keep it to a generic error only.
I leave that up to you.

Regards,

Jan


JanLabrie:
My bad. Fixing.

Issue #104, opened 2020-05-29
Feature: Class path filtering


clemi2408:
Hi Victools-Team, 

as i saw that Feature in the gravitee plugin, i would love to have a filter functionality based on GLOB syntax to include and exclude classes:
https://github.com/gravitee-io/json-schema-generator-maven-plugin

Example:
> Generate all classes ending by ``Schema`` would be written as:
> ```xml
> <include>*Schema.class</include>
> ```

CarstenWickner:
Hi @clemi2408,

I've created PR #132 to address this feature request.
The proposed solution adds support for glob like patterns in the existing `<classNames>` and `<packageNames>` elements and introduces a new `<excludeClassNames>` element counter-part.

Your particular example would be possible, but would look slightly differently:
> Generate all classes ending by Schema would be written as:
> ```xml
> <classNames>**Schema</classNames>
> ```

CarstenWickner:
This has been included in the [v4.14.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.14.0) release.

PR #111, opened 2020-06-09
102 - Other way of selecting classes from a package

JanLabrie:


jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -187,21 +186,13 @@ private void generateSchema(Class<?> schemaClass) throws MojoExecutionException
...
-        Set<Class<?>> subTypes;
-        try {
-            subTypes = reflections.getSubTypesOf(Object.class);
-        } catch (ReflectionsException reflectionsException) {
-            if (reflectionsException.getMessage().contains("Scanner SubTypesScanner was not configured")) {
-                throw new MojoExecutionException("Could not locate any classes in package " + packageName, reflectionsException);
-            } else {
-                throw new MojoExecutionException("Unknown problem when locating classes in package " + packageName, reflectionsException);
+        Reflections reflections = new Reflections("", new SubTypesScanner(false), this.getClassLoader());
+        Set<String> classes = reflections.getAllTypes();
```
CarstenWickner, line 190: Do you have a rough feeling of what this means for the performance?
Would it make sense to only look-up all types once and provide them as input to the method for each `<packageName>` entry in the configuration, e.g.
- `generateSchemaForPackage(Set<String> classesOnClasspath, String packageName)`

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -187,21 +186,13 @@ private void generateSchema(Class<?> schemaClass) throws MojoExecutionException
...
+        for (String className : classes) {
+            if (className.startsWith(packageName)) {
```
CarstenWickner, line 192: I suspect false positives with this, when you have a `<packageName>com.my.app</packageName>` also picking up types from neighbouring packages like `com.my.appservice.Neighbour`.

Seeing this complete collection of all types here, I'm wondering whether this would be a good opportunity to address GH-104 by introducing some kind of glob matching.
For that, having the concept of including and excluding certain patterns would be helpful instead of the strict separation into `<className>` and `<packageName>`. The main challenge I see here would be backward-compatibility, but I guess we could keep them side-by-side for a while.

Issue #110, opened 2020-06-09
feature: support convert class name to snake_case json key


shijinkui:
for example:
`class AbcTech {...}`
convert to 
`{"abc_tech": }`

CarstenWickner:
Hi @shijinkui,

It should be straightforward to introduce something like that for definition keys. Will just need to make sure it works for various naming strategies then.

Some items need to be considered:
- If this is configurable it should still be enforced to be URI compatible (as demanded by the JSON Schema Specification or even stricter when adhering to the OpenAPI Specification). Currently that is achieved by replacing certain characters (like `<`/`>`/`,`) and removing others. For the Snake Case in particular that shouldn’t be an issue.
- When nested classes are encountered, the declaring class‘ name is added as prefix with a `$` symbol as separator. E.g. `Map$Entry`. Is there some special handling needed here? Should it be possible to not add this prefix?
- What about duplicate names then? When encountering multiple classes with the same in different packages or in other scenarios where different custom definitions are applied for the same type, the definition keys currently get an automatic counter separated by a dash, e.g. `MyClass-1`/`MyClass-2` or here then `my_class-1`/`my_class-2`. Would it be necessary for this suffix to be configurable as well? And if yes, what about cases where that leads to name clashes again?

The above are reasons for this not having been implemented yet.
I‘m open for any ideas/suggestions on the above.

JanLabrie:
I also would like to have extended support in this area. In my case it is about changing the name because the name of the Java class is not good enough to be used in the schema. 
The classes I have are named *Bean, which is the implementation of an interface. 

For example 
```
class StepBean implements Step {
```
should be lead to the schema 
```
  "definitions" : {
    "Step" : {
      "type" : "object",
      "properties" : {
      ....
```
I can do some work in this area. In my project I already made an implementation for my need. We can see whether this is generic enough.

CarstenWickner:
HI @shijinkui,

I've created an implementation proposal for this in GH-113, introducing a new configuration option for a `SchemaDefinitionNamingStrategy`, which gives you (almost) full control over what names/keys are used in the `definitions`/`$defs`. For compatibility's sake any illegal characters are replaced/removed afterwards.

For the Snake Case in particular you could use the Jackson `PropertyNamingStrategy` to avoid having to define complicated rules from scratch:
```java
PropertyNamingStrategyBase jacksonSnakeCase = new PropertyNamingStrategy.SnakeCaseStrategy();
[...]

PR #113, opened 2020-06-12
feat: introduce configurable naming strategy for definition names

CarstenWickner:
Hi @JanLabrie,

On GH-110 you mentioned that you already tackled the topic of adjusting the names/keys of subschemas in the `definitions`/`$defs`.
I'd appreciate if you could have a look at my proposal here.

I was aiming for flexibility while preserving the contract of the JSON Schema Specification (and OpenAPI Specification) regarding the allowed characters in this schema keys.

What do you reckon?


JanLabrie:
Hi Carsten.

I will have a look this weekend.

Jan.

Op vr 12 jun. 2020 23:21 schreef Carsten Wickner <notifications@github.com>:

CarstenWickner:
Merging without releasing for now. If there are any concerns, they can be tackled in a separate PR then.

Issue #115, opened 2020-06-18
How to access "parent's" field that holds the reference to the current field


jochenberger:
Sorry, I have to ask another question. That's what you get for being so helpful. ;-)
I'm trying to write a module that consumes Swagger v3 annotations.
I have the following class structure (more or less)
```java
interface IReference {
  String getName();
}

class Reference<T> extends IReference {

}
@Schema(description= "the foo's person")
interface PersonReference extends IReference {
  @Schema(description = "the person's name")
  String getName();
}

class Foo {
  @Schema(implementation = PersonReference.class)
  private Reference<Person> person;
}
```
In my module, I want to resolve the description for `Foo.person.name`, but in the description resolver, I only get a reference to the `name` field in the `Reference` class. I somehow need to resolve the `@Schema#implementation()` from there but I can't seem to access the `Foo.person` field from there.
```java
class MyModule {
  .... 
  fieldConfigPart.withDescriptionResolver(member -> {
    // member is the "name" field in Reference which does not have a @Schema annotation
    // how can I access the "name" field from PersonReference here?
    // I can only move up to Reference<Person> via member.getDeclaringType
  }
}
```
Does that make sense?

CarstenWickner:
Hi @jochenberger,

Sounds like you want something like this:
```java
configBuilder.forFields()
        .withTargetTypeOverridesResolver(field -> {
            Schema annotation = field.getAnnotation(Schema.class);
            if (annotation == null || annotation.schema() == Void.class) {
                return null;
            }
            return Collections.singletonList(field.getContext().resolve(annotation.schema()));
        });
```

This should cause the field to be treated as the right type (i.e. `PersonReference` here).

jochenberger:
Oh, I guess I should have checked whether there are any WiP branches already. I guess I'll dump my custom implementation and take yours instead. Can I somehow help getting that ready to be included?
One thing I noticed is that the methods in `PersonReference` are apparently not being picked up as fields. I need to enable `GETTER_METHODS` for them to be included. Which again requires me to rewrite property names.
Also, your Swagger2Module does not pick up the description from `@Schema` annotations in classes referenced via `implementation` (`"the foo's person"`) in my example.

CarstenWickner:
I'd appreciate any help on this. So maybe checkout/clone the repository including that branch and continue extending its functionality. I'd love to receive a Pull Request from you.

[...]

PR #116, opened 2020-06-19
[WiP] Consider all available Schema annotations when reading annotation values

jochenberger:
Follow-up for #115 
We'll need to fix the formatting and add tests

Missing:
- [ ] fix formatting
- [ ] extract title and description from @Schema annotations on types
- [ ] add tests


jochenberger:
I'm afraid I'm not able to use the checkstyle plugin in Eclipse. Isn't there anything I can run from the command line to apply the rules?

jochenberger:
CheckStyle complains about the class name: `Type name 'Swagger2Module' must match pattern '^[A-Z][a-zA-Z]*[1-9]?$'.` Any suggestions?

CarstenWickner:
Either we disable this checkstyle rule for that particular class or simply rename to something like `SwaggerTwoModule`.

I'm sorry checkstyle doesn't work properly for you in Eclipse, but don't worry. I'll just merge your changes on the feature branch and fix the formatting then.

Issue #76, opened 2020-04-16
Allow for integration in OpenAPI 3.1 


CarstenWickner:
The motivation for this enhancement is this great [blog post from Phil Sturgeon](https://apisyouwonthate.com/blog/openapi-v31-and-json-schema-2019-09), describing how OpenAPI has been a subset and a superset of JSON Schema at the same time (i.e. only parts of the JSON Schema keywords were supported, while additional ones were added on top).

With the upcoming OpenAPI version 3.1, this is supposed to change insofar as that it will be only a superset of JSON Schema Draft 2019-09 (i.e. supporting all its keywords, but adding some more on top of it).   
He also advocates for the API-first approach, but that is mostly applicable to new projects and not to (big) existing ones that would've to undergo extensive (manual) refactoring to produce an appropriate OpenAPI description out of the existing codebase that doesn't have the relatively new `@OpenApi` annotations everywhere.

## Goal
The overall goal here would be to allow for the flexible generation of an OpenAPI definition from a codebase that has little to no `@OpenAPI` annotations on it yet. Based on that, the existing codebase could then be automatically replaced/regenerated from the OpenAPI definition. Et voilá: welcome in the API-first world. From here on out, you can make changes directly in the OpenAPI definition and have those reflected in your code.

The migration complexity would then lie in the configuration of the `jsonschema-generator` to pick-up as much accurate information as possible and include it in the generated models/schemas in the OpenAPI definition to minimize the information loss. The `jsonschema-generator` would then only be a one-time migration tool – assuming you don't need a separate generation of JSON Schemas besides what is included in the OpenAPI definition then.

### First steps: `jsonschema-generator` enhancements
- Allow for the schema generation in the context of the OpenAPI format.
   - [x] Allow for multiple entry/main types to be provided one-after-the-other and only in the very end producing a common list of `definitions`/`$defs`. #84
   - [x] Allow for an alias for `$defs` to be provided, i.e. to allow for `"$ref": "#/components/schemas/YourType"` instead of the currently unchangeable `"$ref": "#/$defs/YourType"`. #84
   - [x] Apply stricter naming rules for definition keys. #84
   - [x] Cater for additional OpenAPI attributes (like `"format": "int32"`) to be set via standard option. #136
   - [x] Create a new `jsonschema-module-swagger-2` to pick-up model/schema related information provided via the respective OpenAPI `@Schema` annotations (for those cases, where parts of the codebase already have them or where a separate JSON Schema is being generated). #117

[...]

PR #117, opened 2020-06-20
Introduce: Module for Swagger 2.x

CarstenWickner:
In the scope of GH-76 and even independently of that, it'd be convenient for out-of-the-box support of Swagger 2.x (OpenAPI) via a new module.

----

For now this is a draft, including changes from @jochenberger and myself.


jsonschema-module-swagger-2/src/test/java/com/github/victools/jsonschema/module/swagger2/IntegrationTest.java
```diff
@@ -0,0 +1,137 @@
...
+            return this.name;
+        }
+    }
+
+    class Person {
+
+    }
+
+    @Schema(description = "the foo's person")
+    class PersonReference extends Reference<Person> {
```
jochenberger, line 123: In my real use-case, this would also be an interface, i.e. `interface PersonReference extends IReference`. That breaks the properties resolution because unfortunately the getter in the interface won't be detected as such.

jochenberger:
What's left to do for this to go in? Is there anything I can help with?

CarstenWickner:
Hi @jochenberger,

I‘d say:
- [ ] fix formatting
- [ ] add some more tests
- [ ] consider more aspects in `forTypesInGeneral()` as well
- [ ] at least briefly investigate how to pick-up annotations from extended interfaces
- [ ] allow to derive fields from methods (already working on this one)

I didn’t have much time recently. Maybe I’ll get to it over the upcoming weekend.

If you have the capacity, you could look into points 3 and 4.

jochenberger:
1 should be done, at least checkstyle doesn't complain anymore.
I'm not sure if 4 is an issue at all, I think that it should work. If you're referring to https://github.com/victools/jsonschema-generator/pull/117#discussion_r443346846, the problem is that the target `implementation` is an interface but there is no implementation. The interface is just there for documentary purposes. Specifically, it allows us to override the description to work around https://github.com/swagger-api/swagger-core/issues/3290.
I could have a look at 3 in the next week, I need to figure out which attributes make sense on a type.

CarstenWickner:
Alright. Merging and releasing the initial version of the `Swagger2Module` as release 4.13.0 now

CarstenWickner:
Release [v4.13.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.13.0) is complete and includes this new `Swagger2Module`.

PR #129, opened 2020-07-23
128 - Incomplete classpath on loading the JavaxValidation/SwaggerX mo…

JanLabrie:
Fix on the classpath of the Modules. Made their dependencies not provided anymore. This makes that they will be transitively included in the maven plugin.

But..this is a choice.

@Carsten Please advice if this is inline with your way of thinking.

### DIFF
--- a/jsonschema-maven-plugin/java.header
+++ b/jsonschema-maven-plugin/java.header
@@ -1,0 +1,18 @@
+^/\*$
+^ \* Copyright (\d\d\d\d-)?\d\d\d\d VicTools\.$
+^ \*$
+^ \* Licensed under the Apache License, Version 2\.0 \(the \"License\"\);$
+^ \* you may not use this file except in compliance with the License\.$
+^ \* You may obtain a copy of the License at$
+^ \*$
+^ \*      http://www\.apache\.org/licenses/LICENSE-2\.0$
+^ \*$
+^ \* Unless required by applicable law or agreed to in writing, software$
+^ \* distributed under the License is distributed on an \"AS IS\" BASIS,$
+^ \* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied\.$
+^ \* See the License for the specific language governing permissions and$
+^ \* limitations under the License\.$
+^ \*/$
+^$
+^package com\.github\.victools\.jsonschema\..*;$
+^$
--- a/jsonschema-maven-plugin/pom.xml
+++ b/jsonschema-maven-plugin/pom.xml
@@ -64,14 +64,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-        </dependency>
-        <dependency>
             <groupId>com.github.victools</groupId>
             <artifactId>jsonschema-generator</artifactId>
         </dependency>
--- a/jsonschema-maven-plugin/pom.xml
+++ b/jsonschema-maven-plugin/pom.xml
@@ -30,37 +30,31 @@
         <dependency>
             <groupId>org.apache.maven.plugin-tools</groupId>
             <artifactId>maven-plugin-annotations</artifactId>
-            <version>${maven.tools.version}</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-plugin-api</artifactId>
-            <version>${maven.core.version}</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-core</artifactId>
-            <version>${maven.core.version}</version>
             <scope>provided</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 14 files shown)


jsonschema-maven-plugin/java.header
```diff
@@ -0,0 +1,18 @@
...
+^ \*$
+^ \* Unless required by applicable law or agreed to in writing, software$
+^ \* distributed under the License is distributed on an \"AS IS\" BASIS,$
+^ \* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied\.$
+^ \* See the License for the specific language governing permissions and$
+^ \* limitations under the License\.$
+^ \*/$
+^$
+^package com\.github\.victools\.jsonschema\..*;$
+^$
```
CarstenWickner, line 18: This file should not be necessary. If you build the whole multi-module project and not just the maven plugin, the single `java.header` file in the parent folder will be picked up.

JanLabrie: This will not be picked up when you build from within the child project. It will find the checkstyle.xml from the root, but the referenced java.header is apparently searched for in the directory where the build was started.

JanLabrie: Your line of thinking is clear. I will investigate a different dependency solution.

CarstenWickner: That's what I meant above. You have to build everything, i.e. the whole maven multi-module project.

JanLabrie: > That's what I meant above. You have to build everything, i.e. the whole maven multi-module project.
This is not convenient right? Particularly when you are working on just one module you would like to minimize your build.

But if you prefer not to duplicate the java.header file I can live with that because it will only fail at the end of the build. I just have to surpress my unpleasant feeling when I get the checkstyle error thrown at me. :-)

JanLabrie:
On the real issue:
I am kind of puzzled. Apparently the javax.validation-api is not in the classpath. 
But what wonders me is that the Jackson jars **are** on the classpath. What I see is that they are treated in completely the same way as the Javax Validation jars:
- Both mentioned in the dependencies of their Module and marked as provided.
- Both mentioned in the dependencies of the Maven plugin and marked as provided. (On my local branch)

When they are not marked as provided in the Maven plugin it will actually work. But I would like to understand why the Jackson jars are part of the classpath and the Javax Validation jars are not.

CarstenWickner:
The Jackson libraries are special because these are explicitly included (as not`provided`) through the main generator library. Only the modules expect those dependencies to be `provided`.

JanLabrie:
Here is an update of what I have found:
- The problem comes from the fact that the classpath of a maven plugin does **not** contain the generic dependencies of the project, but only the plugin dependencies.
[Maven Plugin Doc](https://maven.apache.org/guides/mini/guide-maven-classloading.html)
This is not new for us since we addressed this before, in the loading of the classes for generation. That's why a URLClassLoader is constructed consisting of the full classpath, thus including the project dependencies.
- But... The JavaxValidationModule was created with a new statement, and therefore was using the plugin classloader. This will eventually fail because javax.validation/validation-api is not on that classpath because it is marked as provided.
- So... I changed the code to dynamically load the JavaxValidationModule from the constructed URLClassLoader. And unfortunately this did not work out. The module can actually be loaded, but the classloader that is registered on the loaded class is the AppClassLoader. This makes that further loading of classes is not done on the URLClassLoader, and therefore again can not find the javax.validation/validation-api classes.
- The next step I want to try is to run the Mojo in a new thread on which the ContextClassLoader is set to the URLClassLoader. This should make that all the work is done within the proper classloader.
- If that will fail as well I am out of options and we have to decided if we should mark the dependencies as not provided. But let's hope we do not have to do that.

CarstenWickner:
Thanks for looking into this @JanLabrie.
I detest those kinds of classpath issues.

But thinking about the scoping once more, couldn’t we declare the dependencies of all modules again in the plugin `pom.xml` as not `provided`?
- Basically doing what you suggested as workaround already.
- The respective versions are already defined in the parent `pom.xml`‘s `<dependencyManagement>` and would not need to be double-maintained.
- No need for complicated code changes while still allowing the modules themselves to keep their `provided` references.
- Only in the scope of the maven plugin are those dependency versions determined by us then, which should not affect any artefact being built further downstream.

JanLabrie:
Your welcome @CarstenWickner.

I agree with you that we should go for the strategy as you describe. Overcomplicating the code in the Maven plugin does not outweigh the benefit we will get from it.

For me there is one counter intuitive element in the solution being the need to mention the transitive dependencies of the modules in the Maven plugin. In my eyes the Maven plugin does not have an interest in knowing that the JavaxValidation module uses javax.validation/validation-api. That should be encapsulated.

But...I understand that when the module is not used via the Maven plugin, but as a direct dependency in a project, you want to use the dependency definition that is already available in the project. Although we can also have some debate about that.

Anyways...let's move forward. So the summary is:

- The version of all dependencies of the whole project are defined in the `<dependencyManagement>` section of the parent POM.
- Dependencies are marked as provided by default.
- In the Maven plugin we add the transitive dependencies of the modules, and do not mark them as provided.
(It is still possible to overwrite this from a users project by specifying dependencies in the plugin area)

Agree? When you do I will make update the PR.

CarstenWickner:
Yes, sounds good.
Fully agree it can be argued both ways whether a library should include all of its transitive dependencies for ease of use.

In my view, the coupling between the maven plugin and the other modules is acceptable since they are always released together.

PR #197, opened 2021-09-14
Update SchemaBuilderTest.java

rfox12:
Small fix to make `createSingleTypeSchema` and `forMultipleTypes` public (and accessible outside of com.github.victools.jsonschema.generator).  Current test examples work because they are within the same package namespace; but currently other users of jsonschema.generator cannot access these methods.  Awesome tool!

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaBuilder.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaBuilder.java
@@ -50,7 +50,7 @@ public class SchemaBuilder {
      * @param typeParameters optional type parameters (in case of the {@code mainTargetType} being a parameterised type)
      * @return generated JSON Schema
      */
-    static ObjectNode createSingleTypeSchema(SchemaGeneratorConfig config, TypeContext typeContext,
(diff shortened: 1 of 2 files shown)


CarstenWickner:
Hi @rfox12,

Thanks for your compliments and taking the time to create this PR.

However, there is no need for this change. Those two methods are already available through the `SchemaGenerator`:
- `SchemaGenerator.generateSchema()` invokes `SchemaBuilder.createSingleTypeSchema()`
- `SchemaGenerator.buildMultipleSchemaDefinitions()` invokes `SchemaBuilder.forMultipleTypes()`

Please just use the `SchemaGenerator` instead.
I want to be able to change those non-`public` `SchemaBuilder` methods without being concerned about a breaking change.

rfox12:
Got it, makes sense.  I figured out the right path.  The main reason I thought this might have been an oversight is that the [Open API test](https://github.com/victools/jsonschema-generator/blob/master/jsonschema-generator/src/test/java/com/github/victools/jsonschema/generator/SchemaBuilderTest.java) you have uses `forMultipleTypes`... A presumption that tests are often examples (they certainly don't need to be).

CarstenWickner:
Good point. Probably worth adjusting that test then.

rfox12:
I think that's it.  Hopefully you can just hit approve.

PR #199, opened 2021-09-14
Update SchemaBuilderTest.java

rfox12:
Small fix to utilise public APIs.  Decided to make a clean PR


CarstenWickner:
Hm okay, now I understand why the `SchemaBuilder` is being tested. It's the unit test for the `SchemaBuilder` class.
Testing the `SchemaGenerator` instead wouldn't be the atomic test I was going for here.

There are already multiple test classes for the `SchemaGenerator`:
- `SchemaGeneratorSimpleTypesTest`
- `SchemaGeneratorComplexTypesTest`
- `SchemaGeneratorSubtypesTest`
- `SchemaGeneratorCustomDefinitionsTest`
- `SchemaGeneratorAllOfCleanUpTest`

Thanks again for your time, but I'm not convinced that this is an improvement. Maybe adding a comment in the test would be sufficient, pointing out that this is not the public `SchemaGenerator` API.
If it's truly confusing that there is a public `SchemaGenerator` and a non-public `SchemaBuilder`, I guess your initial PR to make the `SchemaBuilder` public is the more appropriate choice after all. Then it wouldn't matter which entry point is used. The `SchemaGenerator` would only continue to exist for backward-compatibility reasons (and because it's the more intuitive entry point due to it more closely matching the library name).

PR #213, opened 2021-12-03
Correctly handle nullable enums

CarstenWickner:
As pointed out in https://github.com/victools/jsonschema-generator/discussions/212, JSON Schema Validators (e.g. the "networknt/json-schema-validator") do not allow `null` values for a schema like the following:
```json
{
  "type": ["string", "null"],
  "enum": ["A", "B", "C"]
}
```
Instead, the schema needs to look like this:
```json
{
  "anyOf": [
    {
      "type": "null"
    },
    {
      "type": "string",
      "enum": ["A", "B", "C"]
    }
  ]
}
```
----

This is already working correctly when the same type is being encountered both in nullable and non-nullable form.
But if a given type only appears in nullable form, the wrong schema is being created.

----

While updating the test accordingly, I also introduced a small change to the default `ObjectMapper` instance, which is being used when no specific one is given in the `SchemaGeneratorConfigBuilder` constructor invocation.
Now, that default `ObjectMapper` avoids scientific notation of numbers, i.e., now should write `10` instead of `1E+1`. This may have no effect, depending on your usage of the `ObjectMapper`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### `jsonschema-generator`
+#### Changed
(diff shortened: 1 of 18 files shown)


CarstenWickner:
Yes, [v4.21.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.21.0) and newer.

PR #248, opened 2022-04-22
toPrettyString README.md

GaetanoPiazzolla:
Using toPrettyString() will provide a nicer output for the first run of the program :)


CarstenWickner (approved):
Fair enough 😅

PR #254, opened 2022-05-14
feat: new Option.SUPPLIER_TYPE_AS_ADDITIONAL_PROPERTIES

hashar:
I am generating schema for Gerrit events which use
com.google.common.base.Supplier (now java.util.function.Supplier), an
example class would be:

 public class CommentAddedEvent extends PatchSetEvent {
   static final String TYPE = "comment-added";
   public Supplier<AccountAttribute> author;
   public Supplier<ApprovalAttribute[]> approvals;
   public String comment;

The `author` and `approvals` were represented with `type: "object"` when
the supplied types `AccountAttribute` and `ApprovalAttribute[]` are way
more complicated.

There is the same issue with `Map<K, V>` which is easily resolved by
using `Option.MAP_VALUES_AS_ADDITIONAL_PROPERTIES`. This patch is merely
a copy paste for `Supplier` type which returns the supplied type.

The feature is not enabled in any of the three defaults.

I have added a few test cases which I believe reflect the intent and a
FAQ entry which might help future users.


hashar:
In my code at https://gerrit.wikimedia.org/r/c/operations/software/gerrit/jsonschemagenerator/+/791647/4/src/main/java/org/wikimedia/gerrit/App.java I went to adapt the FAQ code for `Map` and instead check for `Supplier` and return the first parameter.  It is not that complicated but surely setting this `Option.SUPPLIER_TYPE_AS_ADDITIONAL_PROPERTIES` would have made my afternoon easier :]

CarstenWickner:
Hi @hashar,

Thank you for pull request.
Especially taking the time to not only write the code change but also the appropriate document. Much appreciated.
I'm open to adding an `Option` to handle `Supplier<T>` types.

That being said, I'm not convinced your changes are generating the schema you're after.
Do you have a JSON example for the example event type given above?

Just looking at the Java code, I'd assume the `Supplier` should be treated like an `Optional` (as per `Option.FLATTENED_OPTIONALS`). By that I mean, this class:
```java
public class EventExample {
    public Supplier<String> eventType;
}
```
is likely serialized as this kind of JSON instance:
```json
{
    "eventType": "comment-added"
}
```
I.e., the appropriate schema would be this:
```json
{
    "type": "object",
    "properties": {
        "eventType": {
            "type": "string"
        }
    }
}
```
However (without testing it explicitly) I suspect your code currently produces this instead:
```json
{
    "type": "object",
    "properties": {
        "eventType": {
            "type": "object",
            "additionalProperties": {
                "type": "string"
            }
        }
    }
}
```

You might have taken the wrong existing `Option` as the basis for your change.
Would be great, if you could provide a JSON example for a Java class using `Supplier` as a property type.

hashar:
Thank you for the encouraging words, happy to see you find support for `Supplier` to be a worthwhile addition.

My code kept `type: object` and used `additionalProperties` as you have pointed out. I have looked at the  code for `Option.FLATTENED_OPTIONS` (nice hint thanks) and in my code I went with:

```java
        configBuilder.forFields()
            .withTargetTypeOverridesResolver(field -> {
                return field.getType().isInstanceOf(Supplier.class)
                    ? Collections.singletonList(field.getTypeParameterFor(Supplier.class, 0))
                    : null;
            });
```
That does generate a schema which validate a test json ;)

I will rework the PR entirely and add a test for a class using `Supplier` object in `SchemaGeneratorComplexTypesTest.java`.

Danke Schon.

hashar:
I have entirely redone the pull request to be base my code upon `Option.FLATTENED_OPTIONALS`. I still have to test it though.

hashar:
I am not sure what I have done but it no more recognizes `com.google.common.base.Supplier` objects.  Maybe cause this PR does:
```java
private static boolean isSupplier(ResolvedType type) {
     return type.getErasedType() == Supplier.class;
}
```

Which I guess expect strictly a `Supplier.class` and would dismiss a `com.google.common.base.Supplier`.  The PR should be updated to use `isInstanceOf` like in the app I wrote:
```
+             .withTargetTypeOverridesResolver(field -> {
+                 return field.getType().isInstanceOf(Supplier.class)
```

The "issue" comes from `FlattenedOptionalModule.java` from which I copy pasted, it does a strict equality as well:
```
    private static boolean isOptional(ResolvedType type) {
        return type.getErasedType() == Optional.class;
    }
```

I am not sure what should be the expected behavior ;)

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### `jsonschema-generator`
+#### Added
+- New `Option.FLATTENED_SUPPLIERS` to unwrap the supplied type. `Supplier<T>` would thus be a `type: "T"`.
```
CarstenWickner, line 10: Should also add as "Changed":
- that the new `Option` is being included in the `OptionPreset.PLAIN_JSON` by default
- that the `Optional.FLATTENED_OPTIONALS` also considers subtypes of `Optional` (after refactoring it accordingly)

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/Option.java
```diff
@@ -87,6 +88,12 @@ public enum Option {
      * @see Option#SIMPLIFIED_OPTIONALS
      */
     FLATTENED_OPTIONALS(FlattenedOptionalModule::new, null, Option.SIMPLIFIED_OPTIONALS),
+    /**
+     * Wheter a {@link java.util.function.Supplier Supplier&lt;T&gt;} instance should be treated as nullable value of the wrapped type {@code T}.
```
CarstenWickner, line 92: To me, the `Supplier<T>` does not necessarily convey that it is "nullable", like the `Optional<T>` does.
The `Supplier<T>` is more to enable lazy-loading of particular values, isn't it? Therefore I'd omit the "nullable" part actually.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/FlattenedSupplierModule.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.function.Supplier;
+
+/**
+ * Default module being included if {@code Option.FLATTENED_SUPPLIERS} is enabled.
+ */
+public class FlattenedSupplierModule implements Module {
+
+    private static boolean isSupplier(ResolvedType type) {
+        return type.getErasedType() == Supplier.class;
```
CarstenWickner, line 33: I agree with your proposal to also consider subtypes of `Supplier`.

I doubt it's very common to implement the functional `Supplier` interface, including it in a data structure for which a JSON Schema is generated and actually meaning something else than this.

For those few edge cases, this `Option` can always be disabled and implemented differently.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/FlattenedSupplierModule.java
```diff
@@ -0,0 +1,59 @@
...
+import com.github.victools.jsonschema.generator.Module;
+import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Supplier;
+
+/**
+ * Default module being included if {@code Option.FLATTENED_SUPPLIERS} is enabled.
+ */
+public class FlattenedSupplierModule implements Module {
```
CarstenWickner, line 30: I'd prefer to refactor the `FlattenedOptionalModule` and extract the common generic part of unwrapping a generic's first argument instead of just adding a second specific module, which is largely identical.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/FlattenedSupplierModule.java
```diff
@@ -0,0 +1,59 @@
...
+                ? Collections.singletonList(fieldOrMethod.getTypeParameterFor(Supplier.class, 0))
+                : null;
+    }
+
+    @Override
+    public void applyToConfigBuilder(SchemaGeneratorConfigBuilder builder) {
+        builder.forFields()
+                .withTargetTypeOverridesResolver(this::resolveSupplierComponentType)
+                // need to getDeclaredType() to avoid the above override
+                .withNullableCheck(field -> FlattenedSupplierModule.isSupplier(field.getDeclaredType()) ? Boolean.TRUE : null);
```
CarstenWickner, line 53: As per my other comment above, I'd rather not imply that a `Supplier` makes its wrapped data "nullable". In my view, it may also only be a practical concern of lazy-loading parts of the data object, if and when it's needed.

jsonschema-generator/src/test/java/com/github/victools/jsonschema/generator/SchemaGeneratorComplexTypesTest.java
```diff
@@ -216,15 +217,16 @@ public List<? extends TestClass2<Long>> getNestedLongList() {
...
             return this.class4;
         }
     }
 
-    private static class TestClass4<S, T> {
+    private static class TestClass4<S, T, P> {
 
         private TestClass2<TestClass2<T>> class2OfClass2OfT;
         public Optional<S> optionalS;
+        public Supplier<P> supplierP;
```
CarstenWickner, line 229: Could also just have made this `supplierS` without adding yet another generic. 😉

slate-docs/source/includes/_main-generator.md
```diff
@@ -108,33 +108,41 @@ configBuilder.without(
       <td>Replacing encountered <code>Optional</code> instances as null-able forms of their generic parameter type.</td>
       <td>-</td>
     </tr>
-    <tr><th>#</th><th>Behavior if included</th><th>Behavior if excluded</th></tr>
```
CarstenWickner, line 111: Let's keep the same number of `Option`s between those repetitions of the table header.

slate-docs/source/includes/_main-generator.md
```diff
@@ -108,33 +108,41 @@ configBuilder.without(
...
     <tr>
       <td rowspan="2" style="text-align: right">9</td>
+      <td colspan="2"><code>Option.FLATTENED_SUPPLIERS</code></td>
+    </tr>
+    <tr>
+      <td>Replacing encountered <code>Supplier</code> instances as null-able forms of the results they supply.</td>
```
CarstenWickner, line 116: As per another comment: let's omit the "nullable" part.

CarstenWickner (approved):
Thank you very much for your contribution. 🙇

I'll take care of the minor clean-ups myself after merging this. 😃

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/Option.java
hashar: I barely know Java myself I don't quite understand what a `Supplier<T>` is for, what I found though is that the supplied T might be a nullable Type like `Long`.  So it probably makes sense to have the schema to reflect the property can be `anyOf` of `T` or `null`.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/FlattenedSupplierModule.java
hashar: I looked a bit again this morning.  If I get a `Supplier<Long>` then `field.getDeclaredType()` seems to be a `Long` which thus never match the conditional `FlattenedSupplierModule.isSupplier`.   The field ends up being or not being nullable after the wrapped value (thus for a `Long` it is a nullable).

Or to say it otherwise, whether the Supplier property is nullable depends on the wrapped type. I think that is good, in the classed I have looked at there are indeed use case for which a `null` is emitted.

jsonschema-generator/src/test/java/com/github/victools/jsonschema/generator/SchemaGeneratorComplexTypesTest.java
hashar: I wasn't sure what to do to address your comment, from #255 turns out it is about not having to use yet another argument `P` but instead reuse the existing `S` leading to:
```java
public Supplier<S> supplierS;
```

Thx!

slate-docs/source/includes/_main-generator.md
hashar: Addressed by #255

hashar:
Thank you very much for your attention and specially for the follow up in #255 . It would have taken me ages to address all the newbie mistakes I have done :]

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/Option.java
CarstenWickner: In the module for the `Optional`, the wrapped type is explicitly set to be null-able even if the default behavior is to treat everything else as not null-able.
I merely prefer to let the `Supplier` not enforce it one way or another and rather let the otherwise configured default do its thing. 😉

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/FlattenedSupplierModule.java
CarstenWickner: Yes, as per my other comment: the default behavior will apply now, i.e., one can still generate a schema that treats everything as not null-able by default (for simplicity's sake), and in that case, the `Supplier` is not forcing it to be null-able.
I believe we're on the same page.

CarstenWickner:
No worries. I'm very opinionated when it comes to "my" code, you didn't make many "mistakes" as such. 😅
Thanks again for your contribution. 😃

CarstenWickner:
This has been released as part of [v4.25.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.25.0) now.

PR #273, opened 2022-08-11
Added parameter to continue build if no classes are found

### LINKED ISSUE #272: Maven Plugin Fails if no classes are found matching Glob
The maven plugin fails if no classes are found matching the given glob pattern.
I would prefer this to be optional, as there are scenarios, where this is not desirable.

I have a parent pom, where I defined the plugin to generate the json schemas for every class matching a certain name. I defined it here, as I want to make sure, that it is executed in every child project. However, in some of the projects (and in the parent project specifically) none of the classes matches the glob pattern. Currently this fails the entire build, as the plugin aborts.

Personally I would prefer it, if there was another mechanism to select the classes for which a schema shall be generated (for example by an annotation). But in a first step making this behavior configurable would already help.

Emmeral:
Fixes #272 
(hopefully, could not test, as the maven plugin tests are broken on Windows (see #274 ))

### DIFF
--- a/jsonschema-maven-plugin/README.md
+++ b/jsonschema-maven-plugin/README.md
@@ -59,6 +59,9 @@ The content of each of these elements can be either:
 </configuration>
 ```
 
+By default, the plugin aborts if the glob pattern does not match any class. If this is not desired,
+the `<failIfNoClassesMatch>` property can be set to `false`.
+
 ### Configuring generated file names and locations
 The location where the files will be generated can be specified with the `<schemaFilePath>` element.
 ```xml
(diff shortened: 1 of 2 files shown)


CarstenWickner (changes requested):

CarstenWickner (approved):
Thanks for your contribution.

I'll add the new parameter in the full documentation (under `/slate-docs` and in the `CHANGELOG` accordingly. Perhaps also some unit test then.

CarstenWickner:
Release of [v4.26.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.26.0) including this change is done.

PR #276, opened 2022-08-15
Schema generation based on Annotations and Classpath restriction

Emmeral:
This PR tries to add more filter mechanisms to the classes for which schemas are generated.
First, the possibility for restricting classes using annotations is introduced. The full qualified name of the annotations can be added in the plugins config.
Second the possibility to restrict the classpath where possible matching classes are searched for is added. In the old implementation the classes where always searched for in the runtime classpath of the project. Because of this, also classes in dependencies and transitive dependencies where found and processed. However, not classes of dependencies with  `<scope>provided<scope>`. Also it was not possible to restrict to only generate schemas for classes from the actual project.
Now all possibilities can be achieved by another configuration parameter.

The latter features are not currently tested by any Unit Test as the structure of the tests has to be changed. This is because the classpath and dependencies of the plugin are no longer included when searching for matching classes. This also makes current tests fail . @CarstenWickner I first wanted some feedback  before further investigating how this could be changed. However, I tried running the snapshot build in a project of mine and everything seemed to work.

### DIFF
--- a/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/Classpath.java
+++ b/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/Classpath.java
@@ -1,0 +1,16 @@
+package com.github.victools.jsonschema.plugin.maven;
+
+public enum Classpath {
+    /**
+     * Only classes in the project.
+     */
+    PROJECT_ONLY,
+    /**
+     * Classes from the project and any runtime dependencies.
+     */
+    WITH_RUNTIME_DEPENDENCIES,
+    /**
+     * Classes from the project and any compile time dependencies.
+     */
+    WITH_COMPILE_DEPENDENCIES
+}
--- a/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
+++ b/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
@@ -485,7 +485,8 @@ public class SchemaGeneratorMojo extends AbstractMojo {
                 classPathElements = project.getRuntimeClasspathElements();
                 break;
             case PROJECT_ONLY:
-                classPathElements = Collections.singletonList(project.getBuild().getOutputDirectory());
+                classPathElements = new ArrayList<>();
+                classPathElements.add(project.getBuild().getOutputDirectory());
                 break;
             default:
                 throw new IllegalArgumentException("Classpath " + classpath + " not supported");
--- a/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
+++ b/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
@@ -195,7 +195,7 @@ public class SchemaGeneratorMojo extends AbstractMojo {
         }
 
         boolean classAndPackageEmpty = (this.classNames == null || this.classNames.length == 0)
-            && (this.packageNames == null || this.packageNames.length == 0);
+                && (this.packageNames == null || this.packageNames.length == 0);
 
         if (classAndPackageEmpty && annotations != null && !annotations.isEmpty()) {
             this.getLog().info("Generating JSON Schema for all annotated classes");
--- a/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
+++ b/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
@@ -45,6 +45,8 @@ import java.net.URLClassLoader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
@@ -90,6 +92,18 @@ public class SchemaGeneratorMojo extends AbstractMojo {
      */
     @Parameter(property = "excludeClassNames")
     private String[] excludeClassNames;
+
+    /**
+     * Full name of annotations for whose annotated classes the JSON schema will be generated.
+     */
+    @Parameter(property = "annotations")
+    private List<String> annotations = new ArrayList<>();
+
+    /**
+     * The classpath to look for classes to generate schema files.
+     */
+    @Parameter(property = "classpath", defaultValue = "WITH_RUNTIME_DEPENDENCIES")
+    private Classpath classpath;
 
     /**
      * The directory path where the schema files are generated.
@@ -184,7 +198,7 @@ public class SchemaGeneratorMojo extends AbstractMojo {
      * Generate the JSON schema for the given className.
      *
      * @param classOrPackageName The name or glob pattern of the class or package
-     * @param targetPackage whether the given name or glob pattern refers to a package
+     * @param targetPackage      whether the given name or glob pattern refers to a package
      * @throws MojoExecutionException In case of problems
      */
     private void generateSchema(String classOrPackageName, boolean targetPackage) throws MojoExecutionException {
@@ -238,15 +252,23 @@ public class SchemaGeneratorMojo extends AbstractMojo {
     private List<PotentialSchemaClass> getAllClassNames() {
         if (this.allTypes == null) {
             Scanner subTypeScanner = Scanners.SubTypes.filterResultsBy(c -> true);
-            URLClassLoader urlClassLoader = this.getClassLoader();
             ConfigurationBuilder configBuilder = new ConfigurationBuilder()
-                    .forPackage("", urlClassLoader)
                     .addScanners(subTypeScanner);
-            if (urlClassLoader != null) {
-                configBuilder.addUrls(urlClassLoader.getURLs());
+            if (annotations != null && !annotations.isEmpty()) {
+                configBuilder.addScanners(Scanners.TypesAnnotated);
             }
+
+            configBuilder.addUrls(getUrls(classpath));
+
             Reflections reflections = new Reflections(configBuilder);
-            Stream<PotentialSchemaClass> allTypesStream = reflections.getAll(subTypeScanner)
+            Set<String> allTypesSet;
+            if (annotations != null && !annotations.isEmpty()) {
+                allTypesSet = reflections.get(Scanners.TypesAnnotated.with(annotations));
+            } else {
+                allTypesSet = reflections.getAll(subTypeScanner);
+            }
+
+            Stream<PotentialSchemaClass> allTypesStream = allTypesSet
                     .stream()
                     .map(PotentialSchemaClass::new);
             if (this.excludeClassNames != null && this.excludeClassNames.length > 0) {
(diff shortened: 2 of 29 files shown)


Emmeral:
I think I solved the problem described above.
@CarstenWickner do you mind having a look and provide some feedback?

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -165,26 +179,35 @@ public synchronized void execute() throws MojoExecutionException {
...
+            generateSchema("**/*", false);
+        }
     }
 
     /**
      * Generate the JSON schema for the given className.
      *
      * @param classOrPackageName The name or glob pattern of the class or package
-     * @param targetPackage whether the given name or glob pattern refers to a package
+     * @param targetPackage      whether the given name or glob pattern refers to a package
```
CarstenWickner, line 210: **suggestion (not-blocking)**: Not ideal to make all these whitespace changes as part of this PR.  Would be nice to minimize the changes to what is required. Also in the unit tests.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -165,26 +179,35 @@ public synchronized void execute() throws MojoExecutionException {
...
                 generateSchema(className, false);
+                executedGenerator = true;
             }
         }
 
         if (this.packageNames != null) {
             for (String packageName : this.packageNames) {
                 this.getLog().info("Generating JSON Schema for <packageName>" + packageName + "</packageName>");
                 generateSchema(packageName, true);
+                executedGenerator = true;
```
CarstenWickner, line 196: **suggestion (blocking)**: This kind of variable re-assignment seems unnecessary.
Instead, check whether the `classNames´ or `packageNames` are neither `null` nor empty in one place.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -238,23 +261,32 @@ private void generateSchema(Class<?> schemaClass) throws MojoExecutionException
...
+            configBuilder.addUrls(getUrls(classpath));
+
             Reflections reflections = new Reflections(configBuilder);
-            Stream<PotentialSchemaClass> allTypesStream = reflections.getAll(subTypeScanner)
+            Set<String> allTypesSet;
+            if (annotations != null && !annotations.isEmpty()) {
+                allTypesSet = reflections.get(Scanners.TypesAnnotated.with(annotations));
+            } else {
+                allTypesSet = reflections.getAll(subTypeScanner);
+            }
```
CarstenWickner, line 278: **suggestion (not-blocking)**: I personally prefer avoiding double negation (`if (not x) a else b`) and rather swap the `if`/`else` blocks in such a case (`if (x) b else a`).
Makes the code slightly more readable in my view.
```suggestion
            if (annotations == null || annotations.isEmpty()) {
                allTypesSet = reflections.getAll(subTypeScanner);
            } else {
                allTypesSet = reflections.get(Scanners.TypesAnnotated.with(annotations));
            }
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/Classpath.java
```diff
@@ -0,0 +1,16 @@
+package com.github.victools.jsonschema.plugin.maven;
+
+public enum Classpath {
```
CarstenWickner, line 3: **suggestion**: The name is ambiguous and too generic. How about `ClasspathFilter` or `ClasspathType` instead?

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -434,29 +466,46 @@ private void setModules(SchemaGeneratorConfigBuilder configBuilder) throws MojoE
      */
     private URLClassLoader getClassLoader() {
         if (this.classLoader == null) {
-            List<String> runtimeClasspathElements = null;
-            try {
-                runtimeClasspathElements = project.getRuntimeClasspathElements();
-            } catch (DependencyResolutionRequiredException e) {
-                this.getLog().error("Failed to resolve runtime classpath elements", e);
-            }
+            List<URL> urls = getUrls(Classpath.WITH_COMPILE_DEPENDENCIES);
```
CarstenWickner, line 469: **suggestion (blocking)**: The `classpath` configuration parameter appears to be ignored.
```suggestion
            List<URL> urls = this.getUrls(this.classpath);
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -434,29 +466,46 @@ private void setModules(SchemaGeneratorConfigBuilder configBuilder) throws MojoE
...
+            this.classLoader = new URLClassLoader(urls.toArray(new URL[0]),
```
CarstenWickner, line 470: **nitpick**: Since we know the number of URLs, might as well initialize the array with the correct length right away.
```suggestion
            this.classLoader = new URLClassLoader(urls.toArray(new URL[urls.size()]),
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -434,29 +466,46 @@ private void setModules(SchemaGeneratorConfigBuilder configBuilder) throws MojoE
...
             }
+        } catch (DependencyResolutionRequiredException e) {
+            this.getLog().error("Failed to resolve classpath elements", e);
+            return Collections.emptyList();
         }
 
-        return this.classLoader;
+        List<URL> urls = new ArrayList<>(classPathElements.size());
+        for (int i = 0; i < classPathElements.size(); i++) {
+            String element = classPathElements.get(i);
```
CarstenWickner, line 501: **suggestion (non-blocking)**: no need to do look-up by index anymore, now that we're not re-using the index variable for assigning entries to particular target array positions.
Seems like a left-over from the previous code. 😉
```suggestion
        for (String element : classPathElements) {
```

jsonschema-maven-plugin/src/test/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojoTest.java
```diff
@@ -197,31 +197,131 @@ public void testPackageName() throws Exception {
...
+        File resultFileA = new File(generationLocation, "TestClassA-schema.json");
+        Assertions.assertFalse(resultFileA.exists());
+        resultFileA.deleteOnExit();
+
+        // TestClassB has the correct annotation
+        File resultFileB = new File(generationLocation, "TestClassB-schema.json");
+        Assertions.assertTrue(resultFileB.exists());
+        resultFileB.deleteOnExit();
+
+
```
CarstenWickner, line 251: **nitpick**: one empty line is enough 😛
The same goes for the other occurrences below.
```suggestion

```

CarstenWickner (changes requested):
Thanks for the effort @Emmeral. Looks overall very promising.
Apart from the mostly minor inline comments, I got only three points:
1. Please add unit tests for the `<classpath>` setting  (as far as feasible without adding more dependencies just for the sake of these tests).
2. With the extra `<annotations>` setting, I'm wondering whether one might want to also filter by a particular annotation attribute, e.g., `@ApiModel(hidden = false)` vs. `@ApiModel(hidden = true)`. How would you adjust the configuration then? We don't need to do that now, I'd just like to consider this as a future extension of the future and prefer to have a consistent way of configuring it.
3. We should add similar extra documentation from the `README` also under the ` slate-docs/source/includes/_maven-plugin.md` file. But I can do that (as well as the `CHANGELOG` entry) afterward myself, i.e., can leave as-is.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
Emmeral: I think I removed all unnecessary white-space changes

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
Emmeral: Changed that

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
Emmeral: Changed that

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/Classpath.java
Emmeral: Changed that

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
Emmeral: No this is intentional. It is used when scanning for files using the reflections library. Here the classloader is retrieved and I thought it would be good, if it contained all dependencies. Then custom modules on the classpath can be loaded independent of the setting.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
Emmeral: Changed that

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
Emmeral: Changed that

jsonschema-maven-plugin/src/test/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojoTest.java
Emmeral: I think I removed all

Emmeral:
> 1. Please add unit tests for the `<classpath>` setting  (as far as feasible without adding more dependencies just for the sake of these tests).

I implemented some unit tests but I could not find a way to test the whole mojo with a single test as in the other tests.

> 2. With the extra `<annotations>` setting, I'm wondering whether one might want to also filter by a particular annotation attribute, e.g., `@ApiModel(hidden = false)` vs. `@ApiModel(hidden = true)`. How would you adjust the configuration then? We don't need to do that now, I'd just like to consider this as a future extension of the future and prefer to have a consistent way of configuring it.

I think one would to alter the configuration to have another level, e.g.
```xml
<annotations>
    <annotation>
        <classname>my.company.mypackage.class</classname>
        <properties>
            <hidden>false</hidden>
        </properties>
    </annotation>
</annotations>
```
and then have a List<AnnotationModel> instead of List<String>

Emmeral:
I implemented my suggestion above. It even works if you only configure it with a string by utilizing a default setter

CarstenWickner:
Thank you for the effort @Emmeral. I'll first put this on a feature branch, to add some more documentation and perhaps introduce the annotation property handling right away, if I can get it to work quickly.

PR #279, opened 2022-08-20
feat: jackson module support @JsonTypeInfo.defaultImpl

CarstenWickner:
When the Jackson module's built-in subtype resolution encounters a `@JsonTypeInfo´ annotation with the `defaultImpl` attribute on it, then that particular subtype is not required to include the designated type property.

From a schema point of view, the assumption is, that some other properties will make it distinct from different subtypes, in order to avoid "false positives" (e.g., another subtype where the type property was forgotten). That is then the responsibility of the schema creator.

Resolves #278, where this change was suggested by @eleftherias.


CarstenWickner:
I can trigger it in a few hours.
Will let you know then.

CarstenWickner:
Release of [v4.26.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.26.0) including this change is done.

PR #282, opened 2022-09-08
feat: java validation support on supplier method

berlam:
You can now add annotations on the supplier method. This is useful for value objects, which consist of one value which resides in a base class.

```java
public class PositiveInteger extends SimpleValueObject<Integer> {

    public PositiveInteger(@NonNull Integer value) {
        super(value);
        isTrue(value >= 0, "value needs to be greater or equal zero");
    }

    @Override
    @Min(0)
    public Integer get() {
        return super.get();
    }
}
```

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/FieldScope.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/FieldScope.java
@@ -25,6 +25,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedParameterizedType;
 import java.lang.reflect.AnnotatedType;
 import java.lang.reflect.Field;
+import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 /**
@@ -112,6 +113,14 @@ public class FieldScope extends MemberScope<ResolvedField, Field> {
      * @return public getter from within the field's declaring class
      */
     private MethodScope doFindGetter() {
+        if (this.getMember().getType().isInstanceOf(Supplier.class)) {
+            ResolvedMethod[] methods = getContext().resolveWithMembers(this.getMember().getType()).getMemberMethods();
+            return Stream.of(methods)
+                    .filter(method -> method.getName().equals("get"))
+                    .findFirst()
+                    .map(method -> this.getContext().createMethodScope(method, this.getDeclaringTypeMembers()))
+                    .orElse(null);
+        }
         String capitalisedFieldName = this.getDeclaredName().substring(0, 1).toUpperCase() + this.getDeclaredName().substring(1);
         String getterName1 = "get" + capitalisedFieldName;
         String getterName2 = "is" + capitalisedFieldName;
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MemberScope.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MemberScope.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 VicTools.
+ * Copyright 2020 VicTools & Sascha Kohlmann
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import com.github.victools.jsonschema.generator.impl.LazyValue;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Member;
 import java.util.Optional;
+import java.util.function.Supplier;
 
 /**
  * Representation of a single introspected field or method.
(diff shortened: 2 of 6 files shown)


jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/FieldScope.java
```diff
@@ -112,6 +113,14 @@ public MethodScope findGetter() {
      * @return public getter from within the field's declaring class
      */
     private MethodScope doFindGetter() {
+        if (this.getMember().getType().isInstanceOf(Supplier.class)) {
```
CarstenWickner, line 116: **suggestion**: While this achieves the specific use-case you're after, it constitutes a breaking change to this library. I suggest making this dependent on the `Option.FLATTENED_SUPPLIERS` at the least.

Plus: the supplier's functional interface implementation is not technically a field's getter. The actual getter could have annotations that may be relevant to consider too, which would get lost here now, wouldn't they?

I'm not sure this is the right place for this.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MemberScope.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 VicTools.
+ * Copyright 2020 VicTools & Sascha Kohlmann
```
CarstenWickner, line 2: Why change the copyright?
This class was originally created in 2019 and while Sascha Kohlmann has provided valuable feedback on multiple occasions and ported the Javax module into the Jakarta module, I don't remember him touching his class here at any time.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MemberScope.java
```diff
@@ -321,14 +322,21 @@ public <A extends Annotation> A getContainerItemAnnotationConsideringFieldAndGet
...
     }
 
+    private boolean isSupplier() {
+        return this.member.getType().isInstanceOf(Supplier.class);
+    }
+
+    private boolean isOptional() {
+        return this.getDeclaredType().getErasedType() == Optional.class
+                && this.getOverriddenType().getErasedType() == this.getDeclaredType().getTypeParameters().get(0).getErasedType();
+    }
```
CarstenWickner, line 338: **suggestion**: While I agree with adding the support for the `Supplier` by default and appreciate the improved readability by extracting these two checks into dedicated methods, I'm not convinced by the method names. The `MemberScope` is neither a `Supplier` nor an `Optional`. It is the member's type that can be either of these two.
The method name should reflect that more clearly, even if that means it becoming more verbose.

Additionally, I'd also include the check of the supplied type against the member's overridden one – same as for `Optional` – and not indiscriminately consider all `Supplier` subtypes.

jsonschema-module-jakarta-validation/src/test/java/com/github/victools/jsonschema/module/jakarta/validation/IntegrationTest.java
```diff
@@ -77,7 +80,7 @@ private static String loadResource(String resourcePath) throws IOException {
         StringBuilder stringBuilder = new StringBuilder();
         try (InputStream inputStream = IntegrationTest.class
                 .getResourceAsStream(resourcePath);
-                Scanner scanner = new Scanner(inputStream, StandardCharsets.UTF_8.name())) {
+             Scanner scanner = new Scanner(inputStream, StandardCharsets.UTF_8.name())) {
```
CarstenWickner, line 83: **suggestion (non-blocking)**: Let's avoid unnecessary whitespace changes in pull requests for features.

jsonschema-module-jakarta-validation/src/test/java/com/github/victools/jsonschema/module/jakarta/validation/IntegrationTest.java
```diff
@@ -128,5 +131,27 @@ static class TestClass {
         @DecimalMin(value = "0", inclusive = false)
         @DecimalMax(value = "1", inclusive = false)
         public double exclusiveRangeDouble;
+
+        @PositiveOrZero
+        public IntegrationTest.TestSupplier supplierPositiveOrZero;
+
+        public TestSupplierWithAnnotation supplierWithAnnotationPositiveOrZero;
+    }
```
CarstenWickner, line 139: **suggestion**:  The tests should also consider a field with type `Supplier<@PositiveOrZero Integer>`.
Also a scenario in which the actual getter method holds the annotation, e.g.
```java
@PositiveOrZero
public Supplier<Integer> getSupplierPositiveOrZeroWithGetter() {
    return this.supplierPositiveOrzeroWithGetter;
}
```

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/FieldScope.java
```diff
@@ -112,6 +113,14 @@ public MethodScope findGetter() {
...
+            ResolvedMethod[] methods = getContext().resolveWithMembers(this.getMember().getType()).getMemberMethods();
+            return Stream.of(methods)
+                    .filter(method -> method.getName().equals("get"))
```
CarstenWickner, line 119: **question**: Collecting all methods only to then filter by name seems excessive. Isn't there an easy way to just look-up the one method we're interested in?

CarstenWickner (changes requested):
Hi @berlam,

Thank you for your input. I've got a few comments/questions but overall appreciate the extended support for the `Supplier` interface.
Although I'm wondering whether the support could be further generalized to consider all `@FunctionalInterface` annotated types, here the single overridden methods expects no arguments and is not `void`. But explicitly supporting `Supplier´ and its subtypes may be good enough already.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/FieldScope.java
CarstenWickner: Since this is just about picking up the annotation on the `get()` method, I'd rather see this as the responsibility of the `Option.FLATTENED_SUPPLIERS`' implementation. It should be part of the reusable type definition rather than having to be inlined for each property.

CarstenWickner: If possible, this should be handled within the `FlattenedWrapperModule`. May need to indicate somehow which method is deemed relevant for this.

PR #299, opened 2022-11-03
Honor inherited annotations

### LINKED ISSUE #297: Inherited javax.validation constraints are ignored
:wave:

It is my understanding that when a getter overrides a method on an interface or a super class, all `javax.validation` constraints on the overridden method also apply to the overriding method (at least that's the behavior I observed).

However, `jsonschema-module-javax-validation` doesn't seem to pick up constraints that are inherited this way. See this failing test:

```java
    @Test
    public void schemaContainsInheritedConstraints() {
        var configBuilder = new SchemaGeneratorConfigBuilder(DRAFT_2019_09)
                .with(new JavaxValidationModule());
        var generator = new SchemaGenerator(configBuilder.build());

        var schema = generator.generateSchema(Book.class);
        var props = (ObjectNode) schema.get("properties");
        var title = (ObjectNode) props.get("title");
        assert title.get("minLength").asInt() == 1;
    }

    class Book implements Publication {
        private String title;

        @Override
        public String getTitle() {
            return title;
        }
    }

    interface Publication {
        @Size(min = 1)
        String getTitle();
    }
```

Is this a bug or am I on the wrong track?

jochenberger:
I tried to fix #297, please advise if this is the correct approach.

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
@@ -23,7 +23,13 @@ import com.github.victools.jsonschema.generator.impl.LazyValue;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedType;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
@@ -171,9 +171,47 @@ public class MethodScope extends MemberScope<ResolvedMethod, Method> {
     public <A extends Annotation> A getAnnotation(Class<A> annotationClass) {
         A annotation = super.getAnnotation(annotationClass);
         if (annotation == null) {
-            annotation = this.getRawMember().getAnnotatedReturnType().getAnnotation(annotationClass);
+            annotation = findGetterCandidates(this.getRawMember()).stream().map(m -> {
+                return Optional.ofNullable(m.getAnnotation(annotationClass))
+                        .orElse(m.getAnnotatedReturnType().getAnnotation(annotationClass));
+            }).filter(Objects::nonNull).findFirst().orElse(null);
         }
         return annotation;
+    }
+
+    private Collection<Method> findGetterCandidates(Method m) {
+        List<Method> candidates = new ArrayList<>();
+        addGetterCandidates(candidates, m.getDeclaringClass(), m.getName());
+        return candidates;
+    }
+
+    private void addGetterCandidates(List<Method> candidates, Class<?> declaringClass,
+            String methodName) {
+        Method getterInClass = findGetterInClass(declaringClass, methodName);
+        if (getterInClass != null) {
+            candidates.add(getterInClass);
+        }
+        for (Class<?> i : declaringClass.getInterfaces()) {
+            Method getterInInterface = findGetterInClass(i, methodName);
+            if (getterInInterface != null) {
+                candidates.add(getterInInterface);
+            }
+
+        }
+        Class<?> superClass = declaringClass.getSuperclass();
+        if (!superClass.equals(Object.class)) {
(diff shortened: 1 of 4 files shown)


jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
```diff
@@ -171,11 +177,49 @@ public boolean isGetter() {
...
+    }
+
+    private void addGetterCandidates(List<Method> candidates, Class<?> declaringClass,
+            String methodName) {
+        Method getterInClass = findGetterInClass(declaringClass, methodName);
+        if (getterInClass != null) {
+            candidates.add(getterInClass);
+        }
+        for (Class<?> i : declaringClass.getInterfaces()) {
+            Method getterInInterface = findGetterInClass(i, methodName);
```
CarstenWickner, line 201: **question**: Would this be covering a method on an extended interface? Maybe we'd have to extend this even further?

Not sure I have considered that everywhere consistently though. Perhaps I should write some more tests to check that. 😉

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
```diff
@@ -171,11 +177,49 @@ public boolean isGetter() {
     public <A extends Annotation> A getAnnotation(Class<A> annotationClass) {
         A annotation = super.getAnnotation(annotationClass);
         if (annotation == null) {
-            annotation = this.getRawMember().getAnnotatedReturnType().getAnnotation(annotationClass);
+            annotation = findGetterCandidates(this.getRawMember()).stream().map(m -> {
+                return Optional.ofNullable(m.getAnnotation(annotationClass))
+                        .orElse(m.getAnnotatedReturnType().getAnnotation(annotationClass));
```
CarstenWickner, line 182: **suggestion**: Two challenges I see here:
> 1. False assumption: When `MethodScope.getAnnotation()` is invoked, all targeted annotations may be inherited.

I'm afraid that doesn't hold true for all types of annotations, i.e., we'd probably have to put this extra logic into the `TypeContext` and allow this to be decided by the caller (on a per-annotation basis) by offering a new/alternative method for it. And then adjust the standard modules where appropriate (e.g. for javax-validation and its jakarta equivalent).

> The "fake container item scope" is not being considered.

E.g., in case of a method returning `List<@NotEmpty String>`, the library is building a schema for the same `MethodScope` twice: first iteration for the `List` and the second iteration for the `String` container item type.  The latter should also be covered in case of overridden methods I guess.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
```diff
@@ -171,11 +177,49 @@ public boolean isGetter() {
...
+            }).filter(Objects::nonNull).findFirst().orElse(null);
         }
         return annotation;
     }
 
+    private Collection<Method> findGetterCandidates(Method m) {
```
CarstenWickner, line 188: **suggestion**: This method is only invoked once with `this.getRawMember()` as its single parameter value.
Therefore, no need to provide any parameter here.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/MethodScope.java
```diff
@@ -171,11 +177,49 @@ public boolean isGetter() {
...
+        List<Method> candidates = new ArrayList<>();
+        addGetterCandidates(candidates, m.getDeclaringClass(), m.getName());
+        return candidates;
+    }
+
+    private void addGetterCandidates(List<Method> candidates, Class<?> declaringClass,
```
CarstenWickner, line 194: **suggestion (non-blocking)**: If `addGetterCandidates()List<Method>, Class<?>, String)` were not `void` but returned the possibly extended`List<Method>`, this code could be shortened a bit.
```suggestion
        return addGetterCandidates(new ArrayList<>(), m.getDeclaringClass(), m.getName());
    }

    private List<Method> addGetterCandidates(List<Method> candidates, Class<?> declaringClass,
```
Might as well remove `findGetterCandidates(Method)`/`findGetterCandidates()` then.

CarstenWickner (changes requested):
Hi @jochenberger,

Thank you very much for your contribution.
I see it works well for your (simple/common) use-case already.

I'd prefer moving all this extra logic into the `TypeContext` class and then extend it further to cover the less common use-cases mentioned in the given comments.

I recently added something similar in the `TypeContext` for type annotations – considering their "hierarchy", i.e., interfaces and super types.

CarstenWickner:
Closing this PR in favor of #344.
Thank you for the input!

PR #309, opened 2023-01-14
feat: support java.time.Period as simple type

vlaxos713:


CarstenWickner (commented):
**question (non-blocking)**: what does a `Period` look like when serialized by Jackson?
Does it match one of the standard `format` values? If yes, then that could be added here as well.

vlaxos713:
It does not match a standard format.

Period when serialized by Jackson looks like the string representation (ISO-8601) standard of a Period. 
From the javadocs of the Period class:
```
Outputs this period as a String, such as P6Y3M1D.
The output will be in the ISO-8601 period format. A zero period will be represented as zero days, 'P0D'.
```

CarstenWickner:
Then "just" need to update the documentation as well. Not sure when I would get to it though.

CarstenWickner (approved):

PR #351, opened 2023-05-11
Fix Option.VALUES_FROM_CONSTANT_FIELDS not triggering

ericbottard:
Previously, Option.VALUES_FROM_CONSTANT_FIELDS could only work when static fields were also considered for exposition as properties.

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/SchemaGeneratorConfigImpl.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/SchemaGeneratorConfigImpl.java
@@ -155,7 +155,7 @@ public class SchemaGeneratorConfigImpl implements SchemaGeneratorConfig {
 
     @Override
     public boolean shouldIncludeStaticFields() {
-        return this.isOptionEnabled(Option.PUBLIC_STATIC_FIELDS) || this.isOptionEnabled(Option.NONPUBLIC_STATIC_FIELDS);
+        return this.isOptionEnabled(Option.PUBLIC_STATIC_FIELDS) || this.isOptionEnabled(Option.NONPUBLIC_STATIC_FIELDS) || this.isOptionEnabled(Option.VALUES_FROM_CONSTANT_FIELDS);
     }
 
     @Override


CarstenWickner:
Hi @ericbottard,

Thank you for your input.

Do I understand it correctly that this is merely to avoid specifically including static fields?
This change alone doesn't do it, right? If neither public nor non-public fields are included, this change doesn't do anything but cause some extra time spent scanning the classes for static fields or am I mistaken?

Perhaps add a unit test to show-case the difference this makes now?

ericbottard:
Hmmm, I'm not sure what you mean "(avoid) specifically including static fields".

This is so that static final fields can be exposed as const constraints, plain and simple. Whether the classes have (non final) static other fields or not. 

Here is an "test-case" (bits omitted) that showcases the issue IMO (by virtue of PLAIN_JSON not considering non-static fields:

```java

...
		configBuilder = new SchemaGeneratorConfigBuilder(SchemaVersion.DRAFT_7, OptionPreset.PLAIN_JSON);

		SchemaGeneratorConfig config = configBuilder.build();
		SchemaGenerator generator = new SchemaGenerator(config);

		return generator.generateSchema(Test.class);
	}

	public static class Test {

		public static final String apiVersion = "foo";

		int regularProperty;

	}
```

gives
```json
{
  "$schema" : "http://json-schema.org/draft-07/schema#",
  "type" : "object",
  "properties" : {
    "regularProperty" : {
      "type" : "integer"
    }
  }
}
```

but changing to FULL_DOCUMENTATION gives

```json
{
  "type" : "object",
  "properties" : {
    "apiVersion" : {
      "type" : "string",
      "const" : "foo"
    },
    "regularProperty" : {
      "type" : "integer"
    }
  }
}
```

Yet, I don't necessarily want to turn static NON FINAL fields as properties of the schema. I just want "const" constraints.

I hope this makes sense

CarstenWickner:
I understand the motivation, but then I'd rather suggest:
- enabling all status fields in general (public or non-public as desired) and
- then configure non-final static fields to be ignored.

ericbottard:
I'm not sure how I would "simply" achieve the second bullet.

In any case, I also assume that each Option should be enabl-able independently of others (unless otherwise documented and implemented), wouldn't you say? Are you saying that VALUES_FROM_CONSTANT_FIELDS requires at least one of PUBLIC_STATIC_FIELDS or NONPUBLIC_STATIC_FIELDS to be enabled?

ericbottard:
Thanks for the clarification and the hint on code snippet, I finally had time to check this today.

This PR can indeed be closed, but may I suggest to 
* clarify in the documentation / javadoc that VALUES_FROM_CONSTANT_FIELDS can't work "on its own",
* since it can't be removed from Preset.PLAIN_JSON b/c of backwards compatibility, maybe document there as well that PUBLIC_STATIC_FIELDS or NONPUBLIC_STATIC_FIELDS needs to be added as well?

In any case, thanks for the quick feedback!

Issue #358, opened 2023-06-05
Support Kotlin annotations


moranlefler:
It seems that the Jackson module ignores Jackson annotations in my Kotlin code, e.g.:
```kotlin
 data class TestNamedProperty( @JsonProperty("my_text")  val text: String)
```

The generator generates a schema with the property name `text` instead of `my_text`. I traced the generation with a debugger, and it seems that the method `getPropertyNameOverrideBasedOnJsonPropertyAnnotation` in the Jackson module does not see the annotation in the Java class that I created from Kotlin using `TestNamedProperty::class.java`.
I suspect that it has to do with how Kotlin and Java annotations differ - https://stackoverflow.com/questions/72559287/how-to-read-kotlin-annotation

I guess that means that `jsonschema-generator` cannot be used from Kotlin, which is a bummer for my use case :(

moranlefler:
It looks like I jumped to a conclusion too quickly as this doesn't seem to be a general problem with Kotlin annotations, as I was able to successfully create a correct schema for a Kotline class like so:
```kotlin
class TestNamedProperty2 {
    @JsonProperty("my_text")
    val text: String = ""
  }
```

So it seems that the issue is specific to Kotlin data classes. I am writing a schema generator that uses `jsonschema-generator` and adds a bit of logic and configurations. Many of its (internal) users use data class so that is a hard requirement. Is this something that I can help fix? any pointers for how to approach this?

moranlefler:
Another relevant reference: https://stackoverflow.com/questions/64118808/java-annotation-cannot-be-found-per-reflection-on-a-kotlin-data-class

CarstenWickner:
Hi @moranlefler,

It's very unfortunate that the annotation goes to the constructor parameter in this scenario.
Currently, those are not taken into account, but they apparently would have to in order to handle data classes correctly.

As starting point, you can extend the `JacksonModule` and override its `getPropertyNameOverrideBasedOnJsonPropertyAnnotation()` method (that's why it has `protected` visibility 😉) to check the constructor parameters as well.
https://github.com/victools/jsonschema-generator/blob/5965fddcd0877a4a07118844dd258d8a12f69dcd/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JacksonModule.java#L194

Preferably, the handling would be generic enough to allow being included via a PR.
Looking forward to your feedback there. 😃

moranlefler:
[...]

PR #359, opened 2023-06-08
KotlinJacksonModule supports JsonProperty annotations on Kotlin data classes

moranlefler:
[victools/jsonschema-generator/issues/358]

I have extended `JacksonModule` and created `KotlinJacksonModule` to support Kotlin data classes.
I have updated `pom.xml` in order to include a Kotlin test class. Please review these changes to make sure it doesn't break anything.
The only functionality currently is to detect `@JsonProperty` annotations on data classes by identifying the appropriate constructor parameter in the compiled class. The functionality only supports the required primary constructor and will ignore any additional constructors.

### DIFF
--- a/jsonschema-module-jackson/pom.xml
+++ b/jsonschema-module-jackson/pom.xml
@@ -51,9 +51,6 @@
     <build>
         <plugins>
             <plugin>
-                <artifactId>maven-compiler-plugin</artifactId>
-            </plugin>
-            <plugin>
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
             <plugin>
--- a/jsonschema-module-jackson/pom.xml
+++ b/jsonschema-module-jackson/pom.xml
@@ -16,6 +16,7 @@
 
     <properties>
         <java.module.name>com.github.victools.jsonschema.module.jackson</java.module.name>
+        <kotlin.version>1.9.0-Beta</kotlin.version>
     </properties>
 
     <dependencies>
@@ -33,6 +34,17 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
             <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-test</artifactId>
+            <version>${kotlin.version}</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
@@ -53,6 +65,19 @@
             <plugin>
                 <groupId>org.moditect</groupId>
                 <artifactId>moditect-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-maven-plugin</artifactId>
+                <version>${kotlin.version}</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <jvmTarget>1.8</jvmTarget>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
             </plugin>
         </plugins>
     </build>
--- a/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
+++ b/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
@@ -1,0 +1,63 @@
+package com.github.victools.jsonschema.module.jackson;
+
+import com.fasterxml.classmate.members.RawField;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.github.victools.jsonschema.generator.MemberScope;
+import kotlin.Metadata;
+
+import java.lang.reflect.Parameter;
+import java.util.List;
+import java.util.OptionalInt;
+
+public class KotlinJacksonModule extends JacksonModule {
+    /**
+     * Look up an alternative name for a member in the constructor parameter list.
+     * When the Kotlin compiler compiles a Kotlin data class, it creates a constructor with a parameter
+     * for each field in the data class. The parameters have generic name such as "arg0", "arg1", etc.
+     * In order to find the appropriate constructor parameter, the method first determines the index of
+     * specified member within its type member list and uses the parameter with the same index.
+     * In order to avoid erroneous overrides, this method verifies that the specified member is indeed of a
+     * Kotlin class.
+     *
+     * @param member field/method to look-up alternative property name for
+     * @return alternative property name or the base class implementation return value
+     */
+    @Override
+    protected String getPropertyNameOverrideBasedOnJsonPropertyAnnotation(MemberScope<?, ?> member) {
+        if (isKotlinType(member)) {
+            OptionalInt memberIndex = getMemberIndex(member);
+            if (memberIndex.isPresent()) {
+                Parameter[] parameters = getConstructorParameters(member);
+                Parameter parameter = parameters[memberIndex.getAsInt()];
+                JsonProperty jsonPropertyAnnotation = parameter.getAnnotation(JsonProperty.class);
+                if (jsonPropertyAnnotation != null) {
+                    String nameOverride = jsonPropertyAnnotation.value();
+                    if (isValidNameOverride(member, nameOverride)) {
+                        return nameOverride;
+                    }
+                }
+            }
+        }
+        return super.getPropertyNameOverrideBasedOnJsonPropertyAnnotation(member);
+
+    }
+
+    private OptionalInt getMemberIndex(MemberScope<?, ?> member) {
+        List<RawField> memberFields = member.getDeclaringType().getMemberFields();
+        for (int i = 0; i < memberFields.size(); i++) {
+            if (memberFields.get(i).getName().equals(member.getName())) {
+                return OptionalInt.of(i);
+            }
+        }
+        return OptionalInt.empty();
+    }
+
+    private static Parameter[] getConstructorParameters(MemberScope<?, ?> member) {
+        return member.getDeclaringType().getConstructors().get(0).getRawMember().getParameters();
+    }
+
+    private static boolean isKotlinType(MemberScope<?, ?> member) {
+        return member.getDeclaringType().getErasedType().isAnnotationPresent(Metadata.class);
+    }
+
+}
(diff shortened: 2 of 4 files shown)


jsonschema-module-jackson/pom.xml
```diff
@@ -34,6 +35,17 @@
             <artifactId>jackson-annotations</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
```
CarstenWickner, line 42: **suggestion**: I'd prefer to avoid adding this extra dependency for all (also non-Kotlin) users.
It is only used to detect Kotlin classes, which can just as easily be steered through a `JacksonOption` and then be applied to all types equally.

At the same time I was just wondering: since the Jackson library supports those annotations on constructor parameters as well (also in Java classes), there is nothing Kotlin-specific about this functionality then.
It just happens to apply to Kotlin data classes too, sure, but the more generic feature to be added here would be to always consider constructor parameters as well as annotations on the field and its getter.

jsonschema-module-jackson/pom.xml
```diff
@@ -54,7 +66,20 @@
...
+                <version>${kotlin.version}</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <jvmTarget>1.8</jvmTarget>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
```
CarstenWickner, line 81: **suggestion**: The Maven compiler plugin is the first one to be mentioned in this `pom.xml`, i.e., this repeated mention is redundant, right?

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+        for (int i = 0; i < memberFields.size(); i++) {
+            if (memberFields.get(i).getName().equals(member.getName())) {
+                return OptionalInt.of(i);
+            }
+        }
+        return OptionalInt.empty();
+    }
+
+    private static Parameter[] getConstructorParameters(MemberScope<?, ?> member) {
+        return member.getDeclaringType().getConstructors().get(0).getRawMember().getParameters();
```
CarstenWickner, line 55: **suggestion (non-blocking)**: Shouldn't we assume that multiple constructors could also exist? Only considering the first might be sufficient for Kotlin data classes, but perhaps more general support of annotations on constructor parameters would be achieved by taking others into account as well?
```suggestion
        return member.getDeclaringType().getConstructors().get(0).getRawMember().getParameters();
```

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private static boolean isKotlinType(MemberScope<?, ?> member) {
+        return member.getDeclaringType().getErasedType().isAnnotationPresent(Metadata.class);
+    }
```
CarstenWickner, line 60: **note**: This check alone doesn't seem worth it to add an extra dependency for all consumers of this library.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+                JsonProperty jsonPropertyAnnotation = parameter.getAnnotation(JsonProperty.class);
+                if (jsonPropertyAnnotation != null) {
+                    String nameOverride = jsonPropertyAnnotation.value();
+                    if (isValidNameOverride(member, nameOverride)) {
+                        return nameOverride;
+                    }
+                }
+            }
+        }
+        return super.getPropertyNameOverrideBasedOnJsonPropertyAnnotation(member);
```
CarstenWickner, line 40: **suggestion**: I'd still favor annotations on the field and its getter method before even considering to check the constructor parameters.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+     * specified member within its type member list and uses the parameter with the same index.
+     * In order to avoid erroneous overrides, this method verifies that the specified member is indeed of a
+     * Kotlin class.
+     *
+     * @param member field/method to look-up alternative property name for
+     * @return alternative property name or the base class implementation return value
+     */
+    @Override
+    protected String getPropertyNameOverrideBasedOnJsonPropertyAnnotation(MemberScope<?, ?> member) {
+        if (isKotlinType(member)) {
```
CarstenWickner, line 26: **suggestion**: As per another comment, let's not limit this to Kotlin types alone.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+     * In order to find the appropriate constructor parameter, the method first determines the index of
+     * specified member within its type member list and uses the parameter with the same index.
+     * In order to avoid erroneous overrides, this method verifies that the specified member is indeed of a
+     * Kotlin class.
+     *
+     * @param member field/method to look-up alternative property name for
+     * @return alternative property name or the base class implementation return value
+     */
+    @Override
+    protected String getPropertyNameOverrideBasedOnJsonPropertyAnnotation(MemberScope<?, ?> member) {
```
CarstenWickner, line 25: **suggestion**: There are other aspects where Jackson annotations are being considered -- not just for the property name overrides. It'd probably be worth it to generalize this a bit more to extend the coverage.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.fasterxml.classmate.members.RawField;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.github.victools.jsonschema.generator.MemberScope;
+import java.lang.reflect.Parameter;
+import java.util.List;
+import java.util.OptionalInt;
+import kotlin.Metadata;
+
+public class KotlinJacksonModule extends JacksonModule {
```
CarstenWickner, line 11: **suggestion**: Let's please not introduce an alternative module just for Kotlin. Instead, we should extend the `JacksonModule`'s capabilities in general and thereby also cover the Kotlin use-case.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Look up an alternative name for a member in the constructor parameter list.
+     * When the Kotlin compiler compiles a Kotlin data class, it creates a constructor with a parameter
+     * for each field in the data class. The parameters have generic name such as "arg0", "arg1", etc.
+     * In order to find the appropriate constructor parameter, the method first determines the index of
+     * specified member within its type member list and uses the parameter with the same index.
+     * In order to avoid erroneous overrides, this method verifies that the specified member is indeed of a
+     * Kotlin class.
```
CarstenWickner, line 19: **note**: I understand your motivation but would prefer to not make this Kotlin specific. Perhaps other JVM languages have similar behavior. It'd be a pity to exclude those without any real need.

CarstenWickner (changes requested):
Thanks a lot for your contribution.
I believe this will be a great addition. As per inline comments though, I'd prefer to not make this strictly Kotlin-specific.
It's a pity though, that the parameter names are not preserved on the implicit constructor of the Kotlin data classes.

jsonschema-module-jackson/pom.xml
moranlefler: In that case, would you prefer checking for the annotation in the appropriate ctor parameter in `FieldScope.getAnnotationConsideringFieldAndGetter`? We can add an `Option` to enable this extra check

jsonschema-module-jackson/pom.xml
moranlefler: According to Kotlin docs (https://kotlinlang.org/docs/maven.html#compile-kotlin-and-java-sources):
> To compile projects that include Kotlin and Java source code, invoke the Kotlin compiler before the Java compiler. In Maven terms it means that kotlin-maven-plugin should be run before maven-compiler-plugin using the following method, making sure that the kotlin plugin comes before the maven-compiler-plugin in your pom.xml file

That's why the additional `maven-compiler-plugin` was added (automatically by the IDE, BTW). It makes the first mention redundant though - I'll remove it.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/KotlinJacksonModule.java
moranlefler: For Kotlin data classes, we can safely assume that all fields are initialized by the implicit ctor in their order in the field list. I'm afraid that might not be the case for other non-data JVM classes. 

e.g.
```java
class CtorTest{
    final int bar;
    @JsonProperty("my_bar")
    final String foo;
    public CtorTest(@JsonProperty("my_foo") int foo, String bar){
        this.bar=foo;
        this.foo=bar;
    }
}
```

How would you suggest we associate between fields and ctor parameters in the general case (non-Kotlin data classes)?

CarstenWickner: @moranlefler I don't think we need to support such a mix of member and constructor parameter annotations.

First of all: it's probably a good idea to put this new feature behind a dedicated `JacksonOption` and then we could start with the assumption that there is one constructor with the same number of parameters as there are fields in the class and that those either match by parameter name or position.

I would have preferred matching by parameter name, but if I understood your code comment correctly, for the Kotlin data class that wouldn't work as they are generically called `arg0`, `arg1`, etc.
Therefore, we need to do it purely based on the index.
If someone else needs a different behavior in the future, we could easily extend this, but this approach should be sufficient for the Kotlin data classes as a start. No need to over-engineer it from the start.
Perhaps nobody uses those annotations on constructor parameters anymore outside of auto-generated constructors such as these. 😃

CarstenWickner:
As per comment on #358, it appears one has to merely indicate where the Java annotation gets applied through an explicit "use site target" (e.g., `@get:JsonProperty`) instead of leaving it up to the Kotline compiler to pick the first (which would be the constructor parameters here).
https://kotlinlang.org/docs/annotations.html#annotation-use-site-targets

With this in place, no additional change is required within the JSON Schema Generator, i.e., I'm closing this PR.
I will later create a new one, wherein this detail will be added to the FAQs in the documentation, to make it more visible in the future.

Issue #384, opened 2023-10-23
Generate schemas in yml format?


takanuva15:
Hi, thanks for this great library! I am using the schema autogeneration example [here](https://github.com/victools/jsonschema-generator/blob/main/jsonschema-examples/src/main/java/com/github/victools/jsonschema/examples/ExternalRefPackageExample.java) which is perfect for the json validation I am doing in my integration tests. I noticed that all the schemas are generated in json format, which is clunkier than the yml schemas that I'm used to. Is it possible to configure the plugin to generate the json-schema files in yaml format? 

For reference, the jackson ObjectMapper I'm using in my code can read yml schema files easily with this configuration:
```
var objectMapper = new ObjectMapper(new YAMLFactory());
```

EDIT: I did some more research and I found that the maven plugin writes the jsonSchema to file in this method within `SchemaGeneratorMojo.java`:
```
    private void writeToFile(JsonNode jsonSchema, File file) throws MojoExecutionException {
        try (FileOutputStream outputStream = new FileOutputStream(file);
                PrintWriter writer = new PrintWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8))) {
            writer.print(jsonSchema.toPrettyString());
        } catch (IOException e) {
            throw new MojoExecutionException("Error: Can not write to file " + file, e);
        }
    }
```
Can we configure this line `writer.print(jsonSchema.toPrettyString());` to instead accept any custom ObjectMapper specified by the user to do the writing? I see that the `SchemaGeneratorConfigBuilder` has an objectmapper associated with it - could we customize that and then use that for the writing so that I can adjust the objectmapper to print out yaml via `objectMapper.writeValueAsString(...)` ?

CarstenWickner:
Hi @takanuva15,

I'm glad the library is of use to you.
I see no reason to prevent this from being generated as YAML files.
Allowing for a custom `ObjectMapper` sounds good. That merely begs the question of how to declare that.
I suppose the YAML support could be a hard-coded option of the plugin.

Either way, I'm open for a Pull Request. 😃
I don't know when I would get to do something about this myself.

takanuva15:
Thx for the quick reply. I'll have to see when I can find some time to look at the library internals for implementation lol. 

[...]

PR #388, opened 2023-10-24
feat: add `withObjectMapper` method to customize serialization of schemas to file

takanuva15:
Hi, I put together this rough PR based on our discussion in #384. 

3 quick questions on this:
1. Assuming the logic looks fine to you, can you clarify which test classes I need to update? (So far I found `SchemaGeneratorConfigBuilderTest.java`).
2. Do I need to update the github.io documentation you have published on the gh-pages branch?
3. Is it worthwhile to add an example within jsonschema-examples demonstrating the customization of the ObjectMapper to print yaml?

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
@@ -17,6 +17,7 @@
 package com.github.victools.jsonschema.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.github.victools.jsonschema.generator.impl.TypeContextFactory;
 import java.lang.reflect.Type;
@@ -74,5 +75,15 @@ public class SchemaGenerator {
      */
     public SchemaBuilder buildMultipleSchemaDefinitions() {
         return SchemaBuilder.forMultipleTypes(this.config, this.typeContext);
+    }
+
+    /**
+     * Returns the {@link ObjectMapper} instance that is associated with the {@link SchemaGeneratorConfig} that was
+     * used to build this {@link SchemaGenerator}.
+     *
+     * @return {@link ObjectMapper}  instance
+     */
+    public ObjectMapper getObjectMapper() {
+        return this.config.getObjectMapper();
     }
 }
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
@@ -78,12 +78,11 @@ public class SchemaGenerator {
     }
 
     /**
-     * Returns the {@link ObjectMapper} instance that is associated with the {@link SchemaGeneratorConfig} that was
-     * used to build this {@link SchemaGenerator}.
+     * Returns the {@link SchemaGeneratorConfig} associated with this {@link SchemaGenerator}.
      *
-     * @return {@link ObjectMapper}  instance
+     * @return a {@link SchemaGeneratorConfig} instance
      */
-    public ObjectMapper getObjectMapper() {
-        return this.config.getObjectMapper();
+    public SchemaGeneratorConfig getConfig() {
+        return this.config;
     }
 }
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
@@ -17,7 +17,6 @@
 package com.github.victools.jsonschema.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.github.victools.jsonschema.generator.impl.TypeContextFactory;
 import java.lang.reflect.Type;
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
@@ -19,6 +19,7 @@ package com.github.victools.jsonschema.generator;
 import com.fasterxml.classmate.AnnotationInclusion;
 import com.fasterxml.jackson.core.json.JsonWriteFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.github.victools.jsonschema.generator.impl.SchemaGeneratorConfigImpl;
 import java.lang.annotation.Annotation;
@@ -41,7 +42,9 @@ public class SchemaGeneratorConfigBuilder {
      * @return default ObjectMapper instance
      */
     private static ObjectMapper createDefaultObjectMapper() {
-        ObjectMapper mapper = new ObjectMapper();
+        ObjectMapper mapper = new ObjectMapper()
+                // enables pretty-printing by default (users can override this by supplying their own mapper)
+                .enable(SerializationFeature.INDENT_OUTPUT);
         mapper.getSerializationConfig()
                 // since version 4.21.0
                 .with(JsonWriteFeature.WRITE_NUMBERS_AS_STRINGS);
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
@@ -43,7 +43,7 @@ public class SchemaGeneratorConfigBuilder {
      */
     private static ObjectMapper createDefaultObjectMapper() {
         ObjectMapper mapper = new ObjectMapper()
-                // enables pretty-printing by default (users can override this by supplying their own mapper)
+                // since version 4.32.0; pretty print by default (can be overridden by supplying explicit mapper)
                 .enable(SerializationFeature.INDENT_OUTPUT);
         mapper.getSerializationConfig()
                 // since version 4.21.0
@@ -311,10 +311,11 @@ public class SchemaGeneratorConfigBuilder {
     }
 
     /**
-     * Register a custom {@link ObjectMapper} to use for serialization of the schema. This can be used to customize
-     * the serialization behavior of the schema (eg serialize to yaml instead of json).
+     * Register a custom {@link ObjectMapper} to create object and array nodes for the JSON structure being
+     * generated. Additionally, it is used to serialize a given schema, e.g., within the standard Maven plugin as
+     * either YAML or JSON.
      *
-     * @param objectMapper the {@link ObjectMapper} to use for serializing schemas
+     * @param objectMapper supplier for object and array nodes for the JSON structure being generated
      * @return this builder instance (for chaining)
      *
      * @since 4.32.0
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
@@ -51,7 +51,7 @@ public class SchemaGeneratorConfigBuilder {
         return mapper;
     }
 
-    private final ObjectMapper objectMapper;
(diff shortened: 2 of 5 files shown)


jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
```diff
@@ -75,4 +76,14 @@ public ObjectNode generateSchema(Type mainTargetType, Type... typeParameters) {
...
+
+    /**
+     * Returns the {@link ObjectMapper} instance that is associated with the {@link SchemaGeneratorConfig} that was
+     * used to build this {@link SchemaGenerator}.
+     *
+     * @return {@link ObjectMapper}  instance
+     */
+    public ObjectMapper getObjectMapper() {
+        return this.config.getObjectMapper();
+    }
```
CarstenWickner, line 88: **suggestion**: Let's add a `SchemaGeneratorConfig getConfig()` getter instead of this delegate.
The Maven plugin can then retrieve the `ObjectMapper` from that, without this being too specific here.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -248,7 +249,7 @@ private void generateSchema(Class<?> schemaClass) throws MojoExecutionException
         JsonNode jsonSchema = getGenerator().generateSchema(schemaClass);
         File file = getSchemaFile(schemaClass);
         this.getLog().info("- Writing schema to file: " + file);
-        this.writeToFile(jsonSchema, file);
+        this.writeToFile(getGenerator().getObjectMapper(), jsonSchema, file);
```
CarstenWickner, line 252: **suggestion**: As per the other comment, let's only expose a `getConfig()` getter on the generator.
```suggestion
        this.writeToFile(getGenerator().getConfig().getObjectMapper(), jsonSchema, file);
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -625,10 +626,10 @@ private void addJacksonModule(SchemaGeneratorConfigBuilder configBuilder, Genera
      * @param file The file to write to
      * @throws MojoExecutionException In case of problems when writing the targeted file
      */
-    private void writeToFile(JsonNode jsonSchema, File file) throws MojoExecutionException {
+    private void writeToFile(ObjectMapper mapper, JsonNode jsonSchema, File file) throws MojoExecutionException {
         try (FileOutputStream outputStream = new FileOutputStream(file);
                 PrintWriter writer = new PrintWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8))) {
-            writer.print(jsonSchema.toPrettyString());
+            writer.print(mapper.writeValueAsString(jsonSchema));
```
CarstenWickner, line 632: **question**: do we need to use the `.writerWithDefaultPrettyPrinter()` in order to still format/indent the output as before? I suppose for YAML, that's irrelevant – or does it then again return JSON?
```suggestion
            writer.print(mapper.writerWithDefaultPrettyPrinter().writeValueAsString(jsonSchema));
```

CarstenWickner (commented):
Thanks for raising this PR 🙇

To answer your questions:
1. The general logic fine, yes. I just added two minor comments. I have no particular test classes in mind, but yes the `SchemaGeneratorConfigBuilderTest` would be good for completeness' sake. The change is not that big I suppose.
2. I can take care of the documentation afterward. For reference though, it's defined as markdown files in the `slate-docs/source/includes` directory of this repository.
3. Examples for the Maven plugin are somewhat tricky in the `jsonschema-examples` scope. An example of using YAML as the output format would be good to include in the FAQs of the documentation I suppose. Feel free to add it yourself, otherwise, I'll do it later.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
takanuva15: Turns out the repo has good unit tests lol - it was failing on my changes because the default ObjectMapper does not pretty-print the json. I did some research, and rather than hardcoding `writerWithDefaultPrettyPrinter()` in the print method, we can toggle the pretty-print option on the ObjectMapper directly:
```
ObjectMapper mapper = new ObjectMapper()
                .enable(SerializationFeature.INDENT_OUTPUT);
```

Now, if someone uses your library and they *don't* want pretty-printing, all they have to do is simply provide their own default ObjectMapper and the pretty-printing won't happen. Thus we knock out 2 problems with this change 🚀 

I've thus updated this PR with that change

takanuva15:
> To answer your questions:
> ...

Thanks for the reply. I've updated the PR and left the documentation untouched. 

This change is really only relevant to the maven plugin, because if a dev is manually making their own schema through the builder, then they'll have access to the root JsonNode and can serialize it however they want. 

In that sense, perhaps the documentation would make more sense to go under `_maven-plugin.md`. In any case, I'll leave it to you to configure it as you like.

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
CarstenWickner: I was considering that too, but that'd be a breaking change for people using a custom `ObjectMapper` already (where no pretty printing is enabled), who still expect the pretty printed output.

Writing this down now, I realise: that would be mostly relevant for users of the Maven plugin, who weren't able to use a custom `ObjectMapper` until now. 😅

I guess it's fine then.

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGenerator.java
```diff
@@ -17,6 +17,7 @@
 package com.github.victools.jsonschema.generator;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
CarstenWickner, line 20: **suggestion**: We can remove the (now again) unused import.
```suggestion
```

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
```diff
@@ -41,7 +42,9 @@ public class SchemaGeneratorConfigBuilder {
      * @return default ObjectMapper instance
      */
     private static ObjectMapper createDefaultObjectMapper() {
-        ObjectMapper mapper = new ObjectMapper();
+        ObjectMapper mapper = new ObjectMapper()
+                // enables pretty-printing by default (users can override this by supplying their own mapper)
```
CarstenWickner, line 46: **suggestion**: Let's include the version number for easier future reference, as for the other settings below.
```suggestion
                // since version 4.32.0; pretty print by default (can be overridden by supplying explicit mapper)
```

jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/SchemaGeneratorConfigBuilder.java
```diff
@@ -306,4 +309,18 @@ public SchemaGeneratorConfigBuilder withAnnotationInclusionOverride(Class<? exte
         this.annotationInclusionOverrides.put(annotationType, override);
         return this;
     }
+
+    /**
+     * Register a custom {@link ObjectMapper} to use for serialization of the schema. This can be used to customize
+     * the serialization behavior of the schema (eg serialize to yaml instead of json).
+     *
+     * @param objectMapper the {@link ObjectMapper} to use for serializing schemas
```
CarstenWickner, line 317: **suggestion**: Let's mention the main purpose of the `ObjectMapper` and not just the scenario mostly relevant to the Maven plugin (as stated on the constructors).
```suggestion
     * Register a custom {@link ObjectMapper} to create object and array nodes for the JSON structure being
     * generated. Additionally, it is used to serialize a given schema, e.g., within the standard Maven plugin as
     * either YAML or JSON.
     *
     * @param objectMapper supplier for object and array nodes for the JSON structure being generated
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -248,7 +249,7 @@ private void generateSchema(Class<?> schemaClass) throws MojoExecutionException
         JsonNode jsonSchema = getGenerator().generateSchema(schemaClass);
         File file = getSchemaFile(schemaClass);
         this.getLog().info("- Writing schema to file: " + file);
-        this.writeToFile(jsonSchema, file);
+        this.writeToFile(getGenerator().getConfig().getObjectMapper(), jsonSchema, file);
```
CarstenWickner, line 252: **suggestion**: I suppose we don't need the extra parameter here and can simply retrieve the `ObjectMapper` within the `writeToFile()` method directly. Both of these methods are `private`, i.e., don't have to cater for additional flexibility in regards to where the `ObjectMapper` is coming from.
```suggestion
        this.writeToFile(jsonSchema, file);
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -621,14 +622,15 @@ private void addJacksonModule(SchemaGeneratorConfigBuilder configBuilder, Genera
     /**
      * Write generated schema to a file.
      *
+     * @param mapper ObjectMapper to use when writing the given schema to the given file
      * @param jsonSchema Generated schema to be written
      * @param file The file to write to
      * @throws MojoExecutionException In case of problems when writing the targeted file
      */
-    private void writeToFile(JsonNode jsonSchema, File file) throws MojoExecutionException {
+    private void writeToFile(ObjectMapper mapper, JsonNode jsonSchema, File file) throws MojoExecutionException {
```
CarstenWickner, line 630: **suggestion**: As per the above comment, there is no real need for the `ObjectMapper` to be passed in as an extra parameter here, since we're getting it from the generally available generator object now.
```suggestion
    private void writeToFile(JsonNode jsonSchema, File file) throws MojoExecutionException {
        ObjectMapper mapper = getGenerator().getConfig().getObjectMapper();
```

jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -621,14 +622,15 @@ private void addJacksonModule(SchemaGeneratorConfigBuilder configBuilder, Genera
     /**
      * Write generated schema to a file.
      *
+     * @param mapper ObjectMapper to use when writing the given schema to the given file
```
CarstenWickner, line 625:
```suggestion
```

jsonschema-maven-plugin/src/test/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojoTest.java
```diff
@@ -75,7 +75,11 @@ public void testGeneration(String testCaseName) throws Exception {
         File referenceFile = new File(testCaseLocation, testCaseName + "-reference.json");
         Assertions.assertTrue(referenceFile.exists());
         Assertions.assertTrue(FileUtils.contentEqualsIgnoreEOL(resultFile, referenceFile, CHARSET_NAME),
-                "Generated schema for " + testCaseName + " is not equal to the expected reference.");
+                "Generated schema for " + testCaseName + " is not equal to the expected reference.\n" +
+                        "Generated: " + FileUtils.readFileToString(resultFile) + "\n" +
+                        "Expected: " + FileUtils.readFileToString(referenceFile));
+        System.out.println(FileUtils.readFileToString(resultFile));
+        System.out.println(FileUtils.readFileToString(referenceFile));
```
CarstenWickner, line 82: **suggestion**: Can remove the sysouts again, now that the tests are succeeding again. 😃
The extra detail in the printout for a failed test is fine, but perhaps add some extra line breaks, to not start on the line saying "Generated:" or "Expected:" but on the one below.
```suggestion
                "Generated schema for " + testCaseName + " is not equal to the expected reference.\n"
                        + "Generated:\n"
                        + FileUtils.readFileToString(resultFile)
                        + "\n----------\n"
                        + "Expected:\n"
                        + FileUtils.readFileToString(referenceFile));
```

CarstenWickner (approved):
Looking good now.
Adding this to the documentation of the Maven plugin is a fair point.
Users are more likely to find it there than in the FAQs.

I'll do that in a separate PR then.
Thank you for this contribution. 😄👍

PR #389, opened 2023-10-25
feat: Maven plugin enhancements

### LINKED ISSUE #383: Enable the maven plugin to read Module class from `src/test/...`?
Hi, I'm using the jsonschema library to generate some json schemas before running my integration tests. All my tests are in a maven module within `module-name/src/test/kotlin/...`. I put my SchemaModule class in `src/test/kotlin/utils/MyModule`, and I can see that it successfully compiles to a .class file during the maven test-compile step. However, the jsonschema maven plugin classloader is unable to find it. After I moved MyModule to `src/main/kotlin/utils/MyModule`, the plugin found the class successfully. 

Can we configure the classloader of the maven plugin to be able to read Module classes out of `src/test/....` too? (I am able to see the MyModule.class file within the `target/test-classes/...` folder that maven generates during compilation?

### LINKED ISSUE #384: Generate schemas in yml format?

CarstenWickner:
Closes #384: Adding documentation after #388 was already merged to provide the feature itself.
Resolves #383: Enable loading custom Module in Maven plugin from test package.

### DIFF
--- a/slate-docs/source/includes/_maven-plugin.md
+++ b/slate-docs/source/includes/_maven-plugin.md
@@ -159,9 +159,9 @@ The given example requires the inclusion of the extra `com.fasterxml.jackson.dat
 </executions>
 ```
 
-When using a custom Module (as mentioned above) for additional configuration options, but don't want to include it among your application code,
+When you're using a custom Module (as mentioned above) for additional configuration options, but don't want to include it among your application code,
 you can either package it as separate artifact and include that as dependency of the plugin (not going into further detail here)
 or the custom Module class can be included in your test packages.  
-When you do the latter, the Maven plugin will by default not be able to load that class, since it won't be compiled yet in the Maven phase the schema generation is being executed at by default.  
+When you do the latter, the Maven plugin will by default not be able to load that class, since it won't be compiled yet in the Maven phase during which the schema generation is being executed.  
 The Maven `compile` phase is when the schema generation gets triggered by default.
 If you want the test classes (including the custom Module) to be available, a later phase (most likely: `test-compile`) needs to be specified.
--- a/slate-docs/source/includes/_maven-plugin.md
+++ b/slate-docs/source/includes/_maven-plugin.md
@@ -122,3 +122,25 @@ Through the `<modules>` tag you can include the standard modules – potentially
 You can also group any kind of configurations into a Module of your own and include it via its full class name.   
 Make sure your custom module is on the classpath (considering the project itself as well as all compile and runtime dependencies) and has a default constructor.    
 It is not possible to configure options for custom modules.
+
+### Altering the format of generated schema files
+
+```java
+public class MavenPluginYamlModule implements Module {
+    @Override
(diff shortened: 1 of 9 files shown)


slate-docs/source/includes/_maven-plugin.md
```diff
@@ -122,3 +123,45 @@ Through the `<modules>` tag you can include the standard modules – potentially
...
+    <execution>
+        <phase>test-compile</phase>
+        <goals>
+            <goal>generate</goal>
+        </goals>
+    </execution>
+</executions>
+```
+
+When using a custom Module (as mentioned above) for additional configuration options, but don't want to include it among your application code,
```
CarstenWickner, line 162:
```suggestion
When you're using a custom Module (as mentioned above) for additional configuration options, but don't want to include it among your application code,
```

slate-docs/source/includes/_maven-plugin.md
```diff
@@ -122,3 +123,45 @@ Through the `<modules>` tag you can include the standard modules – potentially
...
+you can either package it as separate artifact and include that as dependency of the plugin (not going into further detail here)
+or the custom Module class can be included in your test packages.  
+When you do the latter, the Maven plugin will by default not be able to load that class, since it won't be compiled yet in the Maven phase the schema generation is being executed at by default.
```
CarstenWickner, line 165:
```suggestion
When you do the latter, the Maven plugin will by default not be able to load that class, since it won't be compiled yet in the Maven phase during which the schema generation is being executed.  
```

PR #399, opened 2023-11-08
feat: support the "format" standard built-in tags (without the OpenAPI ones)

### LINKED ISSUE #396: Support the "format" string tag without the other OpenAPI format-values
When looking into how to get the `"format"` tag properly formatted following the specs https://json-schema.org/understanding-json-schema/reference/string#format, I found that the only supported option for that is the `Option.EXTRA_OPEN_API_FORMAT_VALUES`. Nevertheless, this introduces`"format"` tags also for properties of numeric types that are not defined in the json-schema specs.

I know that there is a way to implement this with a customized method,  but it would be nice to have support with an `Option` to use the supported built-in formats (https://json-schema.org/understanding-json-schema/reference/string#built-in-formats)  from the json-schema specs (without relying on third-party OpenAPI custom formats).

Is there a possibility to support that feature out-of-the-box with either an option or directly when a version schema supports that built-in format?

*Reproduction*: I realized when I tried to use the `2019-09` schema with a model containing an `UUID` and an `Integer`, expecting to either get a `"format": uuid` with some option (i.e., `Option.ADDITIONAL_FIXED_TYPES`) but the only way was providing the OpenAPI format values that are also adding the `"format": int32` to the `number` and `integer` fields.

magicDGS:
Fixes #396

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/Option.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/Option.java
@@ -286,6 +286,20 @@ public enum Option {
      * @since 4.11.0
      */
     PLAIN_DEFINITION_KEYS(null, null),
+
+    /**
+     * For the "format" attribute, JSON Schema defines various built-in supported values.
+     *
+     * Some of those data-types would be handed if either {@link #ADDITIONAL_FIXED_TYPES}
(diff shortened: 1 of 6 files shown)


CarstenWickner:
Hi @magicDGS,

I like the overall approach
I'll slightly tweak some names, move the paragraph in the documentation, and mention it in the `CHANGELOG`, but I can do that after the merge.
Thank you very much for your contribution. 😄

CarstenWickner (approved):

PR #435, opened 2024-03-08
Fix Jackson JsonUnwrapped to consider inherited members

Nephery:
Members with the `@JsonUnwrapped` annotation were being completely dropped when inherited from a super-class or interface.

This PR fixes the `@JsonUnwrapped` lookup to use `ResolvedMember` instead of `RawMember` so that it would now consider inherited members. I added a concrete example of how this should work to the `IntegrationTest`.


CarstenWickner (approved):
Thank you for the contribution!
I'll include it in the changelog etc. in the next few days then.

Nephery:
@CarstenWickner thanks for accepting the PR. By any chance, could we get a new release with the fix? We have some new schemas that depend on this working. 🙂

CarstenWickner:
Hi @Nephery,

I'll do the release tonight.

CarstenWickner:
Release [v4.34.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.34.0) has been published and will likely take a couple of hours to be fully propagated to the various artifact repositories.

PR #447, opened 2024-05-27
JsonEnumDefaultValue support

tfactor2:
We check the backward compatibility of payloads based on schemas. There is a case that we want to mark as backward compatible, particularly when the enum set is reduced but there was a default enum value set. This will allow developers to fallback to it even if some enum is deleted. Here is the PR to support it in the Jackson module.
 ```java
      enum TestEnumWithJsonEnumDefaultAnnotations {
          A, @JsonEnumDefaultValue B, C
      }
```

->

```json
      "enumValueWithJsonEnumDefaultValueAnnotation": {
            "type": "string",
            "enum": ["A", "B", "C"],
            "default": "B"
        },
```

PS: Not sure about using DefaultResolver in the module.

### DIFF
--- a/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JsonEnumDefaultValueResolver.java
+++ b/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JsonEnumDefaultValueResolver.java
@@ -1,0 +1,20 @@
+package com.github.victools.jsonschema.module.jackson;
+
+import com.fasterxml.jackson.annotation.JsonEnumDefaultValue;
+import com.github.victools.jsonschema.generator.TypeScope;
+import java.lang.reflect.Field;
+import java.util.Arrays;
+
+public class JsonEnumDefaultValueResolver {
+
+    public static String apply(TypeScope typeScope) {
+        if (typeScope.getType().getErasedType().isEnum()) {
+            return Arrays.stream(typeScope.getType().getErasedType().getDeclaredFields())
+                    .filter(enumValue -> enumValue.isAnnotationPresent(JsonEnumDefaultValue.class))
+                    .findFirst()
+                    .map(Field::getName)
+                    .orElse(null);
+        }
+        return null;
+    }
+}
--- a/jsonschema-module-jackson/src/test/resources/com/github/victools/jsonschema/module/jackson/integration-test-result.json
+++ b/jsonschema-module-jackson/src/test/resources/com/github/victools/jsonschema/module/jackson/integration-test-result.json
@@ -54,6 +54,11 @@
             "type": "string",
             "enum": ["entry1", "entry2", "entry3"]
         },
+      "enumValueWithJsonEnumDefaultValueAnnotation": {
+            "type": "string",
+            "enum": ["A", "B", "C"],
+             "default": "B"
+        },
         "fieldWithDescription": {
             "type": "string",
             "description": "field description"
--- a/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JacksonModule.java
+++ b/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JacksonModule.java
@@ -114,6 +114,8 @@ public class JacksonModule implements Module {
 
         applySubtypeResolverToConfigBuilder(generalConfigPart, fieldConfigPart, methodConfigPart);
 
+        fieldConfigPart.withDefaultResolver(JsonEnumDefaultValueResolver::apply);
+
         generalConfigPart.withCustomDefinitionProvider(new JsonUnwrappedDefinitionProvider());
     }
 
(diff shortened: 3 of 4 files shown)


jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JsonEnumDefaultValueResolver.java
```diff
@@ -0,0 +1,20 @@
...
+import com.fasterxml.jackson.annotation.JsonEnumDefaultValue;
+import com.github.victools.jsonschema.generator.TypeScope;
+import java.lang.reflect.Field;
+import java.util.Arrays;
+
+public class JsonEnumDefaultValueResolver {
+
+    public static String apply(TypeScope typeScope) {
+        if (typeScope.getType().getErasedType().isEnum()) {
+            return Arrays.stream(typeScope.getType().getErasedType().getDeclaredFields())
```
CarstenWickner, line 12: **suggestion**: Since `MethodScope.getType()` can return `null` for `void` methods and you're only applying this to fields, I'd suggest making the declared parameter type `FieldScope` instead.
Also `getErasedType()` may return `null`, I believe, if the given type is merely a generic type placeholder.
```suggestion
    public static String apply(FieldScope scope) {
        Class<?> erasedType = scope.getType().getErasedType();
        if (erasedType != null && erasedType.isEnum()) {
            return Arrays.stream(erasedType.getDeclaredFields())
```

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JsonEnumDefaultValueResolver.java
```diff
@@ -0,0 +1,20 @@
+package com.github.victools.jsonschema.module.jackson;
+
+import com.fasterxml.jackson.annotation.JsonEnumDefaultValue;
+import com.github.victools.jsonschema.generator.TypeScope;
```
CarstenWickner, line 4: As per below suggestion, declaring the more specific `FieldScope` instead of `TypeScope` as method parameter type.
```suggestion
import com.github.victools.jsonschema.generator.FieldScope;
```

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JsonEnumDefaultValueResolver.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Arrays;
+
+public class JsonEnumDefaultValueResolver {
+
+    public static String apply(TypeScope typeScope) {
+        if (typeScope.getType().getErasedType().isEnum()) {
+            return Arrays.stream(typeScope.getType().getErasedType().getDeclaredFields())
+                    .filter(enumValue -> enumValue.isAnnotationPresent(JsonEnumDefaultValue.class))
+                    .findFirst()
+                    .map(Field::getName)
```
CarstenWickner, line 15: **suggestion (blocking)**: This is too simplistic.
There are several ways to represent an enum value in the JSON (Schema). The declared name in the code may therefore not be the appropriate one to declare as default here.
We should at least attempt to cover the handful of standard options already supported by the main schema generator and this `JacksonModule` (e.g., considering `@JsonValue` annotations).

jsonschema-module-jackson/src/test/resources/com/github/victools/jsonschema/module/jackson/integration-test-result.json
```diff
@@ -54,6 +54,11 @@
             "type": "string",
             "enum": ["entry1", "entry2", "entry3"]
         },
+      "enumValueWithJsonEnumDefaultValueAnnotation": {
+            "type": "string",
+            "enum": ["A", "B", "C"],
+             "default": "B"
```
CarstenWickner, line 60: **nitpick**: The alignment seems a bit off. 😉

CarstenWickner (changes requested):
Hi @tfactor2,

First of all: Thank you for your contribution!
Generally, I don't mind adding support for this Jackson annotation.

A challenge for such additions, like your proposed `JsonEnumDefaultValueResolver`, is that it is probably working fine in your particular use-case but doesn't cover several other scenarios (namely: the variety of representations for a given enum value – for which there have been surprisingly many questions raised in the context of this project in the past).
Now I'm curious how widely used it is to declare a default enum value on the enum itself and not in the context of a particular field/method where it is used.
I'm wondering because I see two options here:
1. We add some complexity to the `JsonEnumDefaultValueResolver`, accepting the fact that this would never cover _all_ scenarios, but would at least cover those already supported by the schema generator today.
2. Or you leave this piece of straightforward configuration in your own code base, where you can make simplifying assumptions such as: "enum values in JSON match the Java/Kotlin enum values".

Is it worth tackling all that complexity or do you rather keep this on your side and we only add this as another example somewhere in the documentation, where others will still be able to benefit from your contribution?
What do you reckon?

Issue #218, opened 2021-12-17
Module for gson


marcelrv:
First of all, thank you for your generator.. It saves me from maintaining documentation by hand.

I was looking for module based on GSON as my project uses that instead of Jackson.
The property name with gson is defined with the @SerializedName("jsonProperty") annotation.

Is there such a module already available to your knowledge (my google search did not result in any..)

If not, any hint on how to get access to this annotation and update the proprty in the schema. Was trying to use  
`.withTypeAttributeOverride((collectedTypeAttributes, scope, context) ` but sofar without success.

CarstenWickner:
Hi @marcelrv,

there's no standard module for `gson` yet to my knowledge.
However, it should be straightforward to achieve what you want with a "Property Name Override", documented here:
https://victools.github.io/jsonschema-generator/#names-of-fields-methods-in-an-object-39-s-properties
You can easily adapt the given `jackson` example to your `gson` use-case, like this:

```java
configBuilder.forFields()
    .withPropertyNameOverrideResolver(field -> Optional
            .ofNullable(field.getAnnotationConsideringFieldAndGetter(SerializedName.class))
            .map(SerializedName::value).orElse(null));
```
The "Type Attribute Override" is more powerful but also way more complicated if you only want to change the name of a given property. 😉

marcelrv:
Thanks, works like a charm!

maybe one more question in this area.
I was hoping to be able to do something similar with `org.eclipse.jdt.annotation.Nullable` but somehow it does not 'see' the annotation

e.g. my class has some members as nullable, some as NonNull. I would like the Nullable ones as not required.
```
    @SerializedName("friendlyName")
    @Expose
    @Schema(description = "Friendly channel name as seen in the UI")
    private @Nullable String friendlyName;
    @SerializedName("channel")
    @Expose
    private @NonNull String channel;
```

Following the examples I assumed this would work well. But even for `friendlyName` which has @Nullable  it returns null
```
configBuilder.forFields()
    .withRequiredCheck(field -> field.getAnnotationConsideringFieldAndGetter(Nullable.class) == null);
```

marcelrv:
Deepdiving a bit I now understand org.eclipse.jdt.annotation.Nullable is not available  because it is a Retention.CLASS annotated class, not a RetentionPolicy.RUNTIME hence not available at runtime.
Dunno if there is a way to still be able to access it... googling does not give clear way

CarstenWickner:
Hi @marcelrv,

[...]

PR #448, opened 2024-05-27
feature: add GSON module

torrespro:
Based on the conversation here (https://github.com/victools/jsonschema-generator/issues/218) and in our implementation in a internal project I added the Gson Module to the project just in case we want to have an initial support to maybe extend in the future.
It's quite simple just 2 annotations for now, in our project actually the module it's just a class in our code not a full new dependency but maybe in some other context and if more functionality is added a module would make more sense.

### DIFF
--- a/jsonschema-generator-parent/pom.xml
+++ b/jsonschema-generator-parent/pom.xml
@@ -165,6 +165,7 @@
         <version.swagger-2>2.2.5</version.swagger-2>
         <!-- maven plugin runtime dependencies -->
         <version.classgraph>4.8.149</version.classgraph>
+        <version.gson>2.11.0</version.gson>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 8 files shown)


CarstenWickner (commented):
Hi @torrespro,

Thanks for creating the new module.
I appreciate your taking the time to put it in all the right places. 😄
Which makes me somewhat torn here.

For completeness' sake, we'd have to add it to the Maven plugin too and then keep dragging the whole `gson` library around with us everywhere.
I'm not a big fan of the latter, especially if we only do all this for the sake of two simple configurations that were already mentioned on the old issue.
I don't like to discard the effort you've put in already, but my current preference would be to merely add this as an example and/or in the documentation, but not add it to the list of maintained modules.

Or do you foresee any other (more complex) additions to the `GsonModule` that would make this worthwhile?
Or is there another (more lightweight) dependency containing only the relevant annotations?
As-is, it seems overkill to me.

PR #449, opened 2024-05-28
Use JsonValue method type when present.

SimY4:
JsonValue can be set not only on enums but on regular types. And the return type of this method will be used as a serialization target type. 

This PR makes Jackson module aware of JsonValue annotations on regular types.


CarstenWickner (changes requested):
Hi @SimY4,

Thanks a lot for your contribution!
It looks already pretty good.
However, I'd suggest trying out the `SubtypeResolver` instead of the `CustomDefinitionProvider` as it avoids handling certain things.

If that works (as I hope it does), it's only a matter of updating the documentation and changelog to include your great addition. 😃

SimY4:
Hi @CarstenWickner . Sorry for taking so long to reply. 

I'm having hard time to see how I can solve this problem with `SubtypeResolver`

`@JsonValue` doesn't really add a subtype relationship to it's class. It overrides the definition i.e. I can have an Email type in Java that gets converted into a string on an API level. There's no subtype relationship here.

This PR is essentially the way I addressed this issue locally and decided to contribute it back. If you think this is not the right approach, I'm happy to leave it as a ticket in the backlog instead. 👍

Issue #455, opened 2024-07-09
Swagger 2's default and allowableValues do not consider the property's type


Nephery:
For the following Swagger 2 module properties:

> From `@Schema(allowableValues = …)` on fields/methods, derive its `"const"`/`"enum"`.
> From `@Schema(defaultValue = …)` on fields/methods, derive its `"default"`.

The outputted JSON always treats them as Strings, and does not consider the actual type of the property.

For example, if I have the following:

```java
public class TestSchema {
	@Schema(allowableValues = "true")
	public boolean myRestrictedBoolean;
}
```

This would produce a schema like:

```json
{
  "type" : "object",
  "properties" : {
    "myRestrictedBoolean" : {
      "type" : "boolean",
      "const" : "true"
    }
  }
}
```

Instead of `"const" : "true"`, I would have expected this to be `"const" : true` (i.e. not a String value). The same happens for any other primitive types.

CarstenWickner:
Hi @Nephery,

Yeah, that's the downside of the Swagger annotation providing everything as `String` values.
We could add extra smarts to cover some straightforward scenarios such as this one, but to be honest, I'd expect something like the jakarta `@AssertTrue` validation annotation to be present in this case, which allows enforcement and can be picked up without having to parse a `String`.

I'm open to a PR though, if you deem it worthwhile.
I fully agree, that the schema currently being generated there is invalid. The issue can be avoided by not using the `@Schema(allowableValues = "true")` in such a case and/or a fix is required here.

Nephery:
That's fair.

I raised #456 to add support for Jakarta `@AssertTrue` and `@AssertFalse`.

Maybe I'll come back around in the future to try and raise a PR to add some support for primitive types in `@Schema(allowableValues = "true")`. But adding support for Jakarta `@AssertTrue` and `@AssertFalse` does actually solve my immediate issue. So that's good enough for me for now. 🙂

CarstenWickner:
The support for the Jakarta `@AssertTrue`/`@AssertFalse` annotations has been released in [v4.36.0](https://github.com/victools/jsonschema-generator/releases/tag/v4.36.0).

I'm leaving this issue open for now, as the original topic (Swagger annotations) has not yet been resolved.

PR #456, opened 2024-07-11
Populate enum/const from Jakarta AssertTrue and/or AssertFalse annotations

Nephery:
Set `enum`/`const` based on the presence of Jakarta `@AssertTrue` and/or `@AssertFalse`.

See https://github.com/victools/jsonschema-generator/issues/455#issuecomment-2221494969 for background info.

### DIFF
--- a/jsonschema-module-jakarta-validation/README.md
+++ b/jsonschema-module-jakarta-validation/README.md
@@ -13,6 +13,7 @@ Module for the [jsonschema-generator](../jsonschema-generator) – deriving JSON
 7. Populate "pattern" for strings. Based on `@Pattern`/`@Email`, when corresponding `JakartaValidationOption.INCLUDE_PATTERN_EXPRESSIONS` is being provided in constructor.
 8. Populate "minimum"/"exclusiveMinimum" for numbers. Based on `@Min`/`@DecimalMin`/`@Positive`/`@PositiveOrZero`.
 9. Populate "maximum"/"exclusiveMaximum" for numbers. Based on `@Max`/`@DecimalMax`/`@Negative`/`@NegativeOrZero`.
+10. Populate "enum"/"const" for booleans. Based on `@AssertTrue`/`@AssertFalse`.
 
 Schema attributes derived from validation annotations on fields are also applied to their respective getter methods.  
 Schema attributes derived from validation annotations on getter methods are also applied to their associated fields.
--- a/jsonschema-module-jakarta-validation/src/main/java/com/github/victools/jsonschema/module/jakarta/validation/JakartaValidationModule.java
+++ b/jsonschema-module-jakarta-validation/src/main/java/com/github/victools/jsonschema/module/jakarta/validation/JakartaValidationModule.java
@@ -28,6 +28,8 @@ import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder;
 import com.github.victools.jsonschema.generator.SchemaGeneratorConfigPart;
 import com.github.victools.jsonschema.generator.SchemaKeyword;
 import jakarta.validation.Constraint;
+import jakarta.validation.constraints.AssertFalse;
+import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.DecimalMax;
 import jakarta.validation.constraints.DecimalMin;
 import jakarta.validation.constraints.Email;
@@ -45,12 +47,13 @@ import jakarta.validation.constraints.PositiveOrZero;
 import jakarta.validation.constraints.Size;
 import java.lang.annotation.Annotation;
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.ToIntBiFunction;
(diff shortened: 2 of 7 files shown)


jsonschema-module-jakarta-validation/src/test/java/com/github/victools/jsonschema/module/jakarta/validation/IntegrationTest.java
```diff
@@ -134,6 +136,14 @@ static class TestClass {
...
         @DecimalMax(value = "1", inclusive = false)
         public double exclusiveRangeDouble;
+
+        @AssertTrue
+        public boolean trueBoolean;
+        @AssertFalse
+        public boolean falseBoolean;
+        @AssertTrue
+        @AssertFalse
+        public Object allEnumResolvedObject;
```
CarstenWickner, line 146: **question**: Isn't this contradictory? How can something be both `true` and `false` simultaneously?
I'd argue this is an invalid use case and we don't need to support it.

We could avoid some complexity if we didn't allow both values to be returned.
For the scenario where you have an `Object` that you want to treat as `boolean`, we should just set it to `"type": "boolean"` instead – but not based on these annotations.
Is this a relevant use case for you, i.e., do we need to offer a way to indicate this type override or is it okay to just remove that part again here?

jsonschema-module-jakarta-validation/src/main/java/com/github/victools/jsonschema/module/jakarta/validation/JakartaValidationModule.java
```diff
@@ -499,6 +504,27 @@ protected BigDecimal resolveNumberExclusiveMaximum(MemberScope<?, ?> member) {
...
+
+        if (this.getAnnotationFromFieldOrGetter(member, AssertTrue.class, AssertTrue::groups) != null) {
+            values.add(true);
+        }
+
+        if (this.getAnnotationFromFieldOrGetter(member, AssertFalse.class, AssertFalse::groups) != null) {
+            values.add(false);
+        }
+
+        return values.isEmpty() ? null : values;
```
CarstenWickner, line 525: **suggestion**: As per my other comment, I'd deem it an invalid combination if something is supposed to be both `true` and `false` at once.
Therefore, we could adjust this here slightly to reduce the overall complexity.
```suggestion
        List<Object> values;
        if (this.getAnnotationFromFieldOrGetter(member, AssertTrue.class, AssertTrue::groups) != null) {
            values = Collections.singletonList(true);
        } else if (this.getAnnotationFromFieldOrGetter(member, AssertFalse.class, AssertFalse::groups) != null) {
            values = Collections.singletonList(false);
        } else {
            values = null;
        }
        return values;
```
That would have an impact on your excellent unit tests of course.

jsonschema-module-jakarta-validation/README.md
```diff
@@ -13,6 +13,7 @@ Module for the [jsonschema-generator](../jsonschema-generator) – deriving JSON
 7. Populate "pattern" for strings. Based on `@Pattern`/`@Email`, when corresponding `JakartaValidationOption.INCLUDE_PATTERN_EXPRESSIONS` is being provided in constructor.
 8. Populate "minimum"/"exclusiveMinimum" for numbers. Based on `@Min`/`@DecimalMin`/`@Positive`/`@PositiveOrZero`.
 9. Populate "maximum"/"exclusiveMaximum" for numbers. Based on `@Max`/`@DecimalMax`/`@Negative`/`@NegativeOrZero`.
+10. Populate "enum"/"const" for booleans. Based on `@AssertTrue`/`@AssertFalse`.
```
CarstenWickner, line 16: Thanks for adding this to the `README` straightaway.
Should also add this to the [corresponding documentation](https://github.com/victools/jsonschema-generator/blob/main/slate-docs/source/includes/_jakarta-validation-module.md) at the same time.

But I can take care of that afterward, same as the [`CHANGELOG`](https://github.com/victools/jsonschema-generator/blob/main/CHANGELOG.md) and the list of contributors in the parent [`pom.xml`](https://github.com/victools/jsonschema-generator/blob/main/jsonschema-generator-parent/pom.xml).

CarstenWickner (commented):
Thank you very much for your excellent contribution.

Just one question about the validity of encountering both `@AssertTrue` and `@AssertFalse` simultaneously. If that isn't a relevant scenario for your use case, I'm in favor of removing that.
What do you reckon?

jsonschema-module-jakarta-validation/src/test/java/com/github/victools/jsonschema/module/jakarta/validation/IntegrationTest.java
Nephery: Neither of these is actually relevant to my use case.

I just implemented it that way cause I was thinking of a solution for if a user added both `@AssertTrue` and `@AssertFalse`. But I don't feel particularly strongly about what I did, so I'll just make these annotations mutually  exclusive as you suggested.

jsonschema-module-jakarta-validation/src/main/java/com/github/victools/jsonschema/module/jakarta/validation/JakartaValidationModule.java
Nephery: Do you think it'd be better if an exception was thrown if someone tried setting both `@AssertTrue` and `@AssertFalse`? Instead of silently ignoring `@AssertFalse`?

jsonschema-module-jakarta-validation/src/test/java/com/github/victools/jsonschema/module/jakarta/validation/IntegrationTest.java
Nephery: Fixed the tests for not supporting having both `@AssertTrue` and `@AssertFalse`.

jsonschema-module-jakarta-validation/README.md
Nephery: I updated the slate-docs. I'll leave the changelog and contributors to you.

jsonschema-module-jakarta-validation/src/main/java/com/github/victools/jsonschema/module/jakarta/validation/JakartaValidationModule.java
CarstenWickner: Combining `@AssertTrue` and `@AssertFalse` is a programming error, as the validation always fails.
Technically, I'd ignore it then as both are probably meant to be allowed.
However, adding logic for such a "programming error" seems unnecessary.

Therefore it's okay to assume that this error/mistake won't exist (or will be fixed independently) and keep the logic simple here as per my above suggestion.
I think It's not worth adding extra logic (neither to ignore invalid annotations nor throwing an error during the schema generation) if we assume this scenario will never occur.
In case somebody happens to raise this as a bug in the future, I'd be very curious about the justification for having both at the same time. 😉

CarstenWickner (approved):
This looks very good.
I'm merging this to a feature branch, to adjust the Changelog and Contributor list in the same final merge afterward.

Thank you very much!

PR #464, opened 2024-07-29
chore(docs): add example for dependentRequired usage

### LINKED ISSUE #463: Please put an example of the class IfPresentAlsoRequire of the dependentRequired in the github
Hi, how are you?

I would like to thank you by the excelent work that are you doing ...

But can you please put a little example of the class  "IfPresentAlsoRequire.class" ... I have tried a lot to do to implement but i couldn't ...

I need to create a contract with a condition, if the field was determined by an another in the domain class, i need to validate it against  the contract ... Like this example:

with insurance is update contract enum i need to validate if the descriptions was filled, but otherwise i let the flow pass ...

I think with the dependency requires resolver will work well, i tried to implement but i couldn't find more on the internet ...

Please, can you help me?

`withDependentRequiresResolver()` is expecting the names of other properties to be returned, which should be deemed "required", if the property represented by the given field/method is present. The results of all registered resolvers are being combined.

```java
configBuilder.forFields()
    .withDependentRequiresResolver(field -> Optional
        .ofNullable(field.getAnnotationConsideringFieldAndGetter(IfPresentAlsoRequire.class))
        .map(IfPresentAlsoRequire::value)
        .map(Arrays::asList)
        .orElse(null));
configBuilder.forMethods()
    .withDependentRequiresResolver(method -> Optional.ofNullable(method.findGetterField())
        .map(FieldScope::getSchemaPropertyName)
        .map(Collections::singletonList)
        .orElse(null));
```

This link:
https://victools.github.io/jsonschema-generator/#quot-required-quot-keyword

Thanks by reading ...
Sincerely

Rafael

CarstenWickner:
As requested in the issue, I'm adding a complete version of the provided example for configuring the `"dependentRequired"` keyword.

Resolves #463.

### DIFF
--- a/jsonschema-examples/src/main/java/com/github/victools/jsonschema/examples/DependentRequiredExample.java
+++ b/jsonschema-examples/src/main/java/com/github/victools/jsonschema/examples/DependentRequiredExample.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2024 VicTools.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


RafaelRfs (commented):
Thanks a lot Carsten :)

PR #487, opened 2024-10-27
feat: add support for `com.fasterxml.jackson.annotation.JacksonAnnotationsInside` annotation

### LINKED ISSUE #486: Add support for `com.fasterxml.jackson.annotation.JacksonAnnotationsInside` annotation
Jackson provides with this annotation which allows to indicate that an annotation actually contains jackson annotations. Allowing for "combo annotations".

Example: 

```
@JacksonAnnotationsInside
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE)
@JsonTypeName
@JsonTypeInfo(
  use = JsonTypeInfo.Id.NAME,
  property = "__typename"
)
public @interface GraphQLJsonType {
}
```

The current jackson module doesn't account for this annotation when performing lookups for jackson annotations.

smaarn:
Basically the idea is to ensure that, for any annotation lookup, a fallback to use annotations injected via combo annotations is being used.

It turns out that there was already a mechanism to do that in `TypeContext` (or `*Context` interfaces for what it's worth). There were other locations where it wasn't available so I implemented this manually in a dedicated helper class.

Fixes #486

### DIFF
--- a/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
+++ b/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
@@ -28,7 +28,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-final class JacksonHelper {
+final class AnnotationHelper {
 
     static final Predicate<Annotation> JACKSON_ANNOTATIONS_INSIDE_ANNOTATED_FILTER = new JacksonAnnotationsInsideAnnotatedFilter();
 
@@ -43,7 +43,7 @@ final class JacksonHelper {
         }
     }
 
-    private JacksonHelper() {
+    private AnnotationHelper() {
         super();
     }
 
--- a/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
+++ b/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
@@ -30,18 +30,8 @@ import java.util.stream.StreamSupport;
 
 final class AnnotationHelper {
 
-    static final Predicate<Annotation> JACKSON_ANNOTATIONS_INSIDE_ANNOTATED_FILTER = new JacksonAnnotationsInsideAnnotatedFilter();
-
-    private static final class JacksonAnnotationsInsideAnnotatedFilter implements Predicate<Annotation> {
-        JacksonAnnotationsInsideAnnotatedFilter() {
-            super();
-        }
-
-        @Override
-        public boolean test(Annotation annotation) {
-            return annotation.annotationType().isAnnotationPresent(JacksonAnnotationsInside.class);
-        }
-    }
+    static final Predicate<Annotation> JACKSON_ANNOTATIONS_INSIDE_ANNOTATED_FILTER = (annotation) ->
+            annotation.annotationType().isAnnotationPresent(JacksonAnnotationsInside.class);
 
     private AnnotationHelper() {
         super();
--- a/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
+++ b/jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
@@ -20,13 +20,9 @@ import com.fasterxml.classmate.members.ResolvedMember;
 import com.fasterxml.jackson.annotation.JacksonAnnotationsInside;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
 
 final class AnnotationHelper {
 
@@ -50,11 +46,11 @@ final class AnnotationHelper {
      * @return an empty entry if not found
      */
     static <A extends Annotation> Optional<A> resolveAnnotation(ResolvedMember<?> member, Class<A> annotationClass) {
-        final A annotation = member.getAnnotations().get(annotationClass);
-        if (annotation != null) {
-            return Optional.of(annotation);
-        }
-        return resolveNestedAnnotations(StreamSupport.stream(member.getAnnotations().spliterator(), false), annotationClass);
+        return com.github.victools.jsonschema.generator.AnnotationHelper.resolveAnnotation(
+                member,
+                annotationClass,
+                JACKSON_ANNOTATIONS_INSIDE_ANNOTATED_FILTER
+        );
     }
 
     /**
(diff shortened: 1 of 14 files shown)


CarstenWickner:
@smaarn Thank you very much for the effort in creating this PR.

I'm currently busy and unsure when I will get around to properly reviewing this, but I will soon.
Overall, I agree with the approach. At first glance, I'm merely wondering whether it can be generalized further, but first need to wrap my head around it.

jsonschema-module-jackson/src/test/java/com/github/victools/jsonschema/module/jackson/JacksonHelperTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.Optional;
+
+/**
+ * Unit test class dedicated to the validation of {@link JacksonHelper}.
+ *
+ * @author Antoine Malliarakis
```
CarstenWickner, line 17: I'll rather include you in the `jsonschema-generator-parent/pom.xml` list of `contributors` and continue leaving out any `@author` annotations in individual files' javadocs.
Git blame still preserves that information anyway. 😉

smaarn: OMG. I missed that one ! :facepalm: All apologies will remove !

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JacksonHelper.java
```diff
@@ -0,0 +1,107 @@
...
+import java.lang.reflect.AnnotatedElement;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+final class JacksonHelper {
```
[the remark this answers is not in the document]
smaarn: Fixed in https://github.com/victools/jsonschema-generator/pull/487/commits/f03489c54d9ef1ff05611f95570d83a1130f8b25

jsonschema-module-jackson/src/test/java/com/github/victools/jsonschema/module/jackson/JacksonHelperTest.java
smaarn: Fixed in https://github.com/victools/jsonschema-generator/pull/487/commits/f128992e829da119f2167b2e472e6bf8e72aff60

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/CustomEnumDefinitionProvider.java
```diff
@@ -121,7 +121,7 @@ protected ResolvedMethod getJsonValueAnnotatedMethod(ResolvedType javaType, Sche
         ResolvedMethod[] memberMethods = context.getTypeContext().resolveWithMembers(javaType).getMemberMethods();
         Set<ResolvedMethod> jsonValueAnnotatedMethods = Stream.of(memberMethods)
                 .filter(method -> method.getArgumentCount() == 0)
-                .filter(method -> Optional.ofNullable(method.getAnnotations().get(JsonValue.class)).map(JsonValue::value).orElse(false))
+                .filter(method -> JacksonHelper.resolveAnnotation(method.getRawMember(), JsonValue.class).map(JsonValue::value).orElse(false))
```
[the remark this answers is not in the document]
smaarn: Indeed. While we're at it ! (done in https://github.com/victools/jsonschema-generator/pull/487/commits/fadf84a5aae510567f8c024180a70512bdb02338)

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/CustomEnumDefinitionProvider.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 VicTools.
+ * Copyright 2024 VicTools.
```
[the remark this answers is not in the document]
smaarn: Updated copyright mentions. A little bit pedantic, I give you that, but feels like a tribute to the years  ;) (fixed in https://github.com/victools/jsonschema-generator/pull/487/commits/a6cc19e7f439a94a6bb07ee86bf150be6d42da40)

jsonschema-module-jackson/src/test/java/com/github/victools/jsonschema/module/jackson/JacksonHelperTest.java
CarstenWickner: Nothing to apologize for 😄
Credit where credit is due. I'm just in favor of consistency and opted to collect the names of contributors in that `pom.xml` instead of spreading it all over the codebase.

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/JacksonHelper.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final class JacksonAnnotationsInsideAnnotatedFilter implements Predicate<Annotation> {
+        JacksonAnnotationsInsideAnnotatedFilter() {
+            super();
+        }
+
+        @Override
+        public boolean test(Annotation annotation) {
+            return annotation.annotationType().isAnnotationPresent(JacksonAnnotationsInside.class);
+        }
+    }
```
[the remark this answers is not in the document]
smaarn: I must confess I've been traumatized by anonymous class back in the days. Old habits. fixed in https://github.com/victools/jsonschema-generator/pull/487/commits/f78c48f5e268726d97fe28a0bfd0e8d7c32c048a

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private AnnotationHelper() {
+        super();
+    }
+
+    /**
+     * Resolves the specified annotation on the given resolved member and resolve indirect jackson annotations.
+     *
+     * <p>It uses the same algorithm as
+     * {@link com.github.victools.jsonschema.generator.TypeContext#getAnnotationFromList(Class, List, Predicate)}
```
CarstenWickner, line 44: **question**: If it is the same logic as in the `TypeContext`, why redundantly implement again instead of using that existing code?
I mean, that's the whole point of the method being in the `TypeContext` class. To make it re-usable and at the same time allow extending it if need be, since it's _not_ inside a `static` class. This feels like we're locking it down too much (in this module) on top of it being duplication and not allowing other modules to use it at the same time (inviting even more code duplication down the line).

Perhaps move the `AnnotationHelper` class into the main generator, extract that existing logic from the `TypeContext`, and include it here to (leaving the method in the `TypeContext` merely as delegate to this helper class' equivalent method). What do you reckon?

The only jackson-specific thing left is then the `Predicate` at the top.

CarstenWickner (approved):
Thanks again for the contribution.
I like it.

I'll just do a little bit of housekeeping (update the Changelog and "slate-doc" documentation as well as the `contributors` list), but this looks pretty good already.
Not today anymore though. It's too late. 😃

jsonschema-module-jackson/src/main/java/com/github/victools/jsonschema/module/jackson/AnnotationHelper.java
smaarn: To be totally fair that was my initial thought but I figured that the first step was to have a working principle and to make sure I wasn't breaking any Jackson bound logic.
If you're fine with it I would be glad to do it. But maybe tomorrow (late here too 🤭)

smaarn: Done in https://github.com/victools/jsonschema-generator/pull/487/commits/d559d8c11def38a14f39cd34a1cba3dfcf9516a2 and https://github.com/victools/jsonschema-generator/pull/487/commits/e6e53b831031cce73912f9b4c6e628e3191ad4d2

Issue #499, opened 2024-12-06
Duplicate entries in `required`


jochenberger:
I'm creating a schema for my entity with the `JakartaValidationModule` and `Swagger2Module`. I have an interface that has a method annotated with `@Schema(requiredMode = REQUIRED)`. The interface has a sub-interface and there's a class implementing the sub-interface. Neither have additional validation or schema annotations.  
In the schema, the property is contained in the `required` array twice.  
I'm still trying to find out why, please let me know if you need additional information.

jochenberger:
Alright, it's not about the interfaces. Here's a repro:

Here's a repro:
```java
package org.example;

import java.io.IOException;
import java.util.Set;

import com.fasterxml.jackson.databind.node.ObjectNode;
import com.github.victools.jsonschema.generator.Module;
import com.github.victools.jsonschema.generator.Option;
import com.github.victools.jsonschema.generator.OptionPreset;
import com.github.victools.jsonschema.generator.SchemaGenerator;
import com.github.victools.jsonschema.generator.SchemaGeneratorConfig;
import com.github.victools.jsonschema.generator.SchemaGeneratorConfigBuilder;
import com.github.victools.jsonschema.generator.SchemaVersion;
import com.github.victools.jsonschema.module.swagger2.Swagger2Module;

import io.swagger.v3.core.util.Json;
import io.swagger.v3.oas.annotations.media.ArraySchema;
import io.swagger.v3.oas.annotations.media.Schema;
import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;

public class SchemaTest {

    public static void main(String[] args) throws IOException {
        SchemaGenerator schemaGenerator = createSchemaGenerator();
        ObjectNode schema = schemaGenerator.generateSchema(Bar.class);
        Json.pretty().writeValue(System.out, schema);
    }

    private static SchemaGenerator createSchemaGenerator() {
        Module swagger2Module = new Swagger2Module();
        SchemaGeneratorConfigBuilder configBuilder = new SchemaGeneratorConfigBuilder(
                SchemaVersion.DRAFT_2019_09, OptionPreset.PLAIN_JSON).with(swagger2Module).with(
                        Option.INLINE_ALL_SCHEMAS, Option.NONSTATIC_NONVOID_NONGETTER_METHODS,
                        Option.FIELDS_DERIVED_FROM_ARGUMENTFREE_METHODS);

        SchemaGeneratorConfig config = configBuilder.build();
[...]

PR #503, opened 2024-12-10
chore: add a test for #499

jochenberger:


jochenberger:
Do you want me to test that on a different layer? For example starting with an annotated class and using the actual schema generator?

jochenberger:
I'd appreciate if you could apply our formatting/style rules since I've given up the fight with Maven years ago.

PR #508, opened 2025-01-14
feat: Drop JDK 11 support in favor of JDK 17

CarstenWickner:


hbzhou (approved):
looks good to me, may I know what's your planned strategy for supporting JDK17

PR #532, opened 2025-06-12
fix: extract distinct enum values

tearfur:
When using custom `EnumModule`s, sometimes there may be multiple enums resolving to the same value, so make sure they are distinct to reduce the size of the resulting schema.

### DIFF
--- a/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/EnumModule.java
+++ b/jsonschema-generator/src/main/java/com/github/victools/jsonschema/generator/impl/module/EnumModule.java
@@ -126,6 +126,7 @@ public class EnumModule implements Module {
         }
         return Stream.of(erasedType.getEnumConstants())
                 .map(enumConstant -> enumConstantToString.apply((Enum<?>) enumConstant))
+                .distinct()
                 .collect(Collectors.toList());
     }
 


tearfur:
Thanks for merging. Let me give some context on why I found myself needing this.

**tl:dr I had to implement an encoding scheme that consists of a pair of values, and this scheme had gone through multiple iterations over the years, where some codes are kept as-is in the newer iterations. These codes from the different iterations are what's producing the duplicate values.**

Here's some pseudo-code to demonstrate my situation (It's Kotlin because my code is in Kotlin):

```kotlin
enum class SchemeVersion {
    V1,
    V2,
}

interface CodeInterface {
    val code: String
}

enum class FirstCode(
    @JsonValue // com.fasterxml.jackson.annotation.JsonValue
    override val code: String,
    val scheme: SchemeVersion
    val description: String,
) : CodeInterface {
    A_V1("A", SchemeVersion.V1, "Code A, V1"),
    B_V2("B", SchemeVersion.V1, "Code B, V1"),
    A_V2("A", SchemeVersion.V2, "Code A, V2"),
    C_V2("C", SchemeVersion.V2, "Code C, V2"),
}

enum class SecondCode(
    @JsonValue
    override val code: String,
    val scheme: SchemeVersion
    val description: String,
) : CodeInterface {
    X_V1("X", SchemeVersion.V1, "Code X, V1"),
    Y_V2("Y", SchemeVersion.V1, "Code Y, V1"),
    X_V2("X", SchemeVersion.V2, "Code X, V2"),
    Z_V2("Z", SchemeVersion.V2, "Code Z, V2"),
}
```

I want to serialise the code to a JSON object like this (this is `FirstCode.A_V1` + `SecondCode.Y_V1`):

```json
{
  "firstCode": "A",
  "secondCode": "Y",
  "scheme": "V1"
}
```

So I represented them in my entity object like this:

```kotlin
class CodePair(
    val firstCode: FirstCode,
    val secondCode: SecondCode,
    val scheme: SchemeVersion,
)

class Entity (
    // ...
    val code: CodePair,
    // ...
)
```

Now, the "proper" way to define `Entity::code` in the schema would probably be to list all of the possible permutations of valid `CodePair` values:

```json
{
  "type": "object",
  "properties": {
    "firstCode": { "type": "string" }
    "secondCode": { "type": "string" }
    "scheme": { "type": "string" }
  },
  "enum": [
    { "firstCode": "A", "secondCode": "X", "scheme": "V1" },
    { "firstCode": "A", "secondCode": "Y", "scheme": "V1" },
    // ...
  ]
}
```

But this is not possible in this library, which is why I raised #529. And for the meantime I configured my schema builder with a custom `EnumModule` like this:

```kotlin
configBuilder.with(EnumModule { v ->
    if (v is CodeInterface)
        v.code
    else
        v.name
})
```

And that's how I ended up with duplicate values in the schema `enum` array.

PR #551, opened 2025-10-22
feature: Allow Custom Modules to be configured with options

Tiffel:
I wrote a custom module to handle Jacksons @JsonView Annotations (happy to open source this as soon it's polished). I had to hardcode the parameter, because custom modules are not configurable at the moment.

This is my attempt to improve on that limitation


CarstenWickner:
HI @Tiffel 

Thank you for raising this, and I'm looking forward to your PR extending the existing `JacksonModule` with the capability to support `@JsonView` annotations.
Given your example, I'm considering extending this custom module parameter support from your proposed plain list of options to key-value pairs. Thereby better supporting use cases like your `skipFieldsStartingWith` example.

PR #577, opened 2026-04-01
Fix: Numeric properties are now correctly serialized as numbers

cschwich:
This PR fixes an issue where numeric properties (such as minLength) in the generated JSON Schema were incorrectly serialized as strings instead of numbers. The root cause was a misconfiguration of the ObjectMapper in versions < 5.0.0, where the WRITE_NUMBERS_AS_STRING feature was set but not actually applied. With the migration to Jackson 3 (from version 5.0.0), this bug was resolved, and the setting is now effective.

## Changes

- Ensure numeric properties in the schema are serialized as numbers
- Adjust tests to verify the correct behavior

## Fixed issues

- #576


CarstenWickner (commented):
Hi @cschwich 

Thanks for pointing this out and the contribution.

If this affects the generator library in general, shouldn't the tests inside the main library be adjusted (where the change is) rather than on the Maven plugin?

cschwich:
Hi @CarstenWickner ,

just a friendly reminder regarding this PR.

As far as I can tell, it addresses a regression introduced in 5.0.0 where numeric JSON Schema keywords are serialized as strings due to WRITE_NUMBERS_AS_STRINGS being enabled on the default ObjectMapper.

Please let me know if there's anything missing or if you'd like me to adjust something.

Thanks a lot!

PR #582, opened 2026-05-01
Add the skip parameter to the Maven plugin

### LINKED ISSUE #581: [Feature] Add the skip parameter to the Maven plugin
The Maven plugin should support the `skip` configuration parameter, as many others do (e.g. [clean](https://maven.apache.org/plugins/maven-clean-plugin/examples/skipping-clean.html), [help](https://maven.apache.org/plugin-tools/maven-plugin-plugin/helpmojo-mojo.html#skip)), like:

```xml
<configuration>
    <skip>true</skip>
</configuration>
```

This is useful when you want to skip the execution in specific builds, without requiring dedicated Maven profiles. The corresponding user property defined in the Mojo should have a scope like in `jsonschema.skip` to avoid clashes with other plugins, and should take precedence over the static plugin configuration above.

For instance, I'd like to generate schemas only when building a new release version of an artifact, not in regular CI jobs. With the above
configuration I'd be able to skip the generation and have faster CI jobs, while activating it in release jobs overriding it with `-Djsonschema.skip=false`.

giulong:
Fixes #581

### DIFF
--- a/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
+++ b/jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
@@ -103,6 +103,14 @@ public class SchemaGeneratorMojo extends AbstractMojo {
     private List<AnnotationParameter> annotations = new ArrayList<>();
 
     /**
+     * Set this to "true" to skip invoking any goals or reports of the plugin.
+     *
+     * @since 5.1.0
+     */
+    @Parameter(property = "jsonschema.skip", defaultValue = "false")
+    private boolean skip;
+
+    /**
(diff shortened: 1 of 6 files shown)


jsonschema-maven-plugin/src/main/java/com/github/victools/jsonschema/plugin/maven/SchemaGeneratorMojo.java
```diff
@@ -102,6 +102,14 @@ public class SchemaGeneratorMojo extends AbstractMojo {
     @Parameter(property = "annotations")
     private List<AnnotationParameter> annotations = new ArrayList<>();
 
+    /**
+     * Set this to "true" to skip invoking any goals or reports of the plugin.
+     *
+     * @since 5.1.0
```
giulong, line 108: I guess this would be a new minor version

